PR #12, opened 2025-09-25
View configuration, OpenSeadragon defaults and OpenSeadragon animations

jwindhager:
refactored view config
set OpenSeadragon defaults
made some static variables public
optimize viewer animations
fix OpenSeadragon types


jwindhager:
discussed offline: viewer options should be applicable without reinitializing the viewer

some of the viewer options also need to be set on the TiledImage level

PR #18, opened 2025-10-07
Implement GeoJSON shapes data reader

jwindhager:


src/data/loaders/geojson.ts
```diff
@@ -0,0 +1,198 @@
...
+} from "../../model/shapes";
+import { MultiPolygon, Vertex } from "../../types";
+import { ShapesData } from "../shapes";
+import { AbstractShapesDataLoader } from "./base";
+
+export const GEOJSON_SHAPES_DATA_SOURCE = "geojson";
+
+export type GeoJSONShapesDataSource = ShapesDataSource<
+  typeof GEOJSON_SHAPES_DATA_SOURCE
+>;
```
jwindhager:
```suggestion
// eslint-disable-next-line @typescript-eslint/no-empty-object-type
export interface GeoJSONShapesDataSource
  extends ShapesDataSource<typeof GEOJSON_SHAPES_DATA_SOURCE> {}
```

src/data/loaders/geojson.ts
```diff
@@ -0,0 +1,198 @@
...
+  ): Promise<GeoJSONObject> {
+    const { signal } = options;
+    signal?.throwIfAborted();
+    if (this.dataSource.path !== undefined && this.workspace !== null) {
+      const fh = await this.workspace.getFileHandle(this.dataSource.path);
+      signal?.throwIfAborted();
+      const file = await fh.getFile();
+      signal?.throwIfAborted();
+      const text = await file.text();
+      return JSON.parse(text) as GeoJSONObject;
```
jwindhager:
```suggestion
      return JSON.parse(text) as GeoJSONObject;  // TODO validate
```

src/data/loaders/geojson.ts
```diff
@@ -0,0 +1,198 @@
...
+        const polygon = this._coordsToPolygon(coords);
+        multiPolygons.push({ polygons: [polygon] });
+      } else if (type === "MultiPolygon") {
+        const coords = (geometry as GeoJSONMultiPolygon).coordinates;
+        const polygons = coords.map((polygonCoords) =>
+          this._coordsToPolygon(polygonCoords),
+        );
+        multiPolygons.push({ polygons });
+      }
+    };
```
jwindhager: For readability (and performance, in the general case), I would suggest to avoid nested functions

src/data/loaders/geojson.ts
```diff
@@ -0,0 +1,198 @@
...
+    ) {
+      pushFromGeometry(obj as GeoJSONGeometry);
+    }
+
+    return multiPolygons;
+  }
+
+  private _coordsToPolygon(
+    coords: GeoJSONPosition[][],
+  ): MultiPolygon["polygons"][0] {
```
jwindhager: What's `coords`? Shouldn't the return type be just `Polygon`? In general, I think this function needs some TLC

src/data/loaders/geojson.ts
```diff
@@ -0,0 +1,198 @@
...
+    if (this.dataSource.path !== undefined && this.workspace !== null) {
+      const fh = await this.workspace.getFileHandle(this.dataSource.path);
+      signal?.throwIfAborted();
+      const file = await fh.getFile();
+      signal?.throwIfAborted();
+      const text = await file.text();
+      return JSON.parse(text) as GeoJSONObject;
+    }
+    if (this.dataSource.url !== undefined) {
+      const resp = await fetch(this.dataSource.url);
```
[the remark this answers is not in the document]
cavenel: Good question, we can keep fetching and parsing the full json for now, and if it becomes a bottleneck in the future (or if json files are too big), then we add streaming,
https://www.npmjs.com/package/@streamparser/json-whatwg seems well maintained, could be an option.

src/data/loaders/geojson.ts
```diff
@@ -0,0 +1,198 @@
...
+      if (!resp.ok) {
+        throw new Error(
+          `Failed to fetch GeoJSON: ${resp.status} ${resp.statusText}`,
+        );
+      }
+      const text = await resp.text();
+      return JSON.parse(text) as GeoJSONObject;
```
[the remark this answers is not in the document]
jwindhager: Nothing yet :) would just leave the TODO for now, can think about this in the future?

PR #52, opened 2026-01-22
refactor: "constant" source type, premultiply point sizes

jwindhager:
switch to constant source type, render up to 2048 points objects, revert to premultiplying point sizes


jwindhager:
@copilot whats up? where are your comments?

PR #94, opened 2026-03-26
Add project settings for drawOptions

HenrikeW:

This PR closes [#237](https://scilifelab.atlassian.net/browse/TMAP-237?issueKey=TMAP-237&subProduct=jira-software)

# Description

- adds setters for the drawOptions
- adds UI for the draw options inside the project settings dialogue

# Screenshot

image

# Testing

- go to the landing page and click on "Show project settings"
- expect to see the dialogue shown in the screenshot
- make sure it's usable with mouse and keyboard

# Additional comments

- The drawing tool isn't implemented yet, so you won't see any effect of changing the draw options. If you want to have an indication that they are saved to the application state, you can add `console.log` statements in the setter function, like this:
```
  setDrawOptions: (options) => {
    console.log("Setting draw options:", options);
    set((draft) => {
      draft.drawOptions = { ...draft.drawOptions, ...options };
    });
    console.log("New draw options:", get().drawOptions);
  },
  ```

- The connected issue also mentions the `viewerOptions`. We decided to postpone those to a later issue.

### DIFF
--- a/apps/tissuumaps/src/components/panels/ProjectPanel/ProjectSettingsDialog.tsx
+++ b/apps/tissuumaps/src/components/panels/ProjectPanel/ProjectSettingsDialog.tsx
@@ -86,8 +86,7 @@ export function ProjectSettingsDialog() {
               }
             />
             <FieldDescription className="text-xs text-muted-foreground">
-              Number of scanlines for shape rasterization (higher = more
-              accurate but slower)
+              Number of scanlines for shape rasterization
             </FieldDescription>
           </Field>
         </div>
--- a/apps/tissuumaps/src/components/panels/ProjectPanel/ProjectSettingsDialog.tsx
+++ b/apps/tissuumaps/src/components/panels/ProjectPanel/ProjectSettingsDialog.tsx
@@ -1,4 +1,97 @@
+import { Input } from "@/components/ui/input";
+
+import { useTissUUmaps } from "../../../store";
+import {
+  Field,
+  FieldControl,
+  FieldDescription,
+  FieldLabel,
+} from "../../common/field";
+
 export function ProjectSettingsDialog() {
-  // TODO
-  return <></>;
+  const drawOptions = useTissUUmaps((state) => state.drawOptions);
+  const setDrawOptions = useTissUUmaps((state) => state.setDrawOptions);
+
+  return (
+    <form onSubmit={(e) => e.preventDefault()} aria-label="Project settings">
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: apps/tissuumaps/src/components/panels/ProjectPanel/ProjectSettingsDialog.tsx [hunk 1]
```diff
@@ -1,4 +1,97 @@
...
+                        parseInt(e.target.value, 10) || 1,
+                      ),
+                    })
+                  }
+                />
+              }
+            />
+            <FieldDescription className="text-xs text-muted-foreground">
+              Number of scanlines for shape rasterization (higher = more
+              accurate but slower)
```
jwindhager: I do realize this comes from the TSDoc comment, but I'm not so sure the statement in brackets is actually accurate 🙈 Need to double-check this. In the meantime, I suggest to simply remove it from the UI:

```suggestion
              Number of scanlines for shape rasterization
```

jwindhager (approved):
LGTM!

Regarding your additional comment:
> The drawing tool isn't implemented yet, so you won't see any effect of changing the draw options.

Actually, the `DrawOptions` don't have anything to do with "interactive shapes drawing" we discussed offline, but they control the WebGL-powered rendering of points and shapes - sorry for the unclear naming! Changing the shape stroke width or the number of scanlines for shapes rendering already works (can be verified by changing them while shapes are loaded in the viewer). Changing the global point size factor currently doesn't seem to work as expected, which is a bug outside of the scope of this PR.

This PR highlighted a few things:
- The docstring here is potentially outdated: https://github.com/TissUUmaps/TissUUmaps4/blob/ca88f162411df44ae18271048c92d436feca285c/packages/%40tissuumaps-core/src/model/types.ts#L122
- Apparently, the global point size factor currently doesn't work / trigger a re-render
- Since we will soon introduce shapes drawing, `DrawOptions` is a confusing name and should be renamed to `WebGLOptions` or `RenderOptions`.

All of these are low priority and outside of the scope of this PR. I will create tasks in the backlog for this, so that we don't forget about them.

apps/tissuumaps/src/components/panels/ProjectPanel/ProjectSettingsDialog.tsx
HenrikeW: Alright, changed it now.

jwindhager:
@all-contributors please add @HenrikeW for code

PR #109, opened 2026-04-07
Fix: image settings do not synchronize

cavenel:
**The bug**: In OpenSeadragonController.ts:366, when _createOrUpdateTiledImages finds an existing tiled image state that matches a new ref, it calls _updateTiledImage(tiledImageState) — but tiledImageState.ref still points to the old ref from the previous synchronize call. So _calculateOpacity() reads the old opacity/visibility values, and since they haven't changed from the tiled image state's perspective, nothing gets updated. 

**The fix**: Set tiledImageState.ref = ref before doing any update logic, so the new opacity/visibility values are used when _calculateOpacity is called.


jwindhager (approved):
LGTM

The initial PR description is missing a description of the hide->pan->show bug, but that one is fixed now as well.

PR #111, opened 2026-04-08
feat: add buttons for adding data objects

cavenel:
## Type of change

## List of changes made

- Added "Add" buttons to all data panels (Images, Labels, Points, Shapes, Tables) to allow users to create new items directly from the UI
- Created new `AddDataObjectDialog` component for configurable data source creation with name input and source type selection
- Improved z-index handling for select and combobox dropdowns (increased from z-10/default to z-60) to ensure they display correctly within modals
- Added scrollbar support to right panels by applying `overflow-y-auto` styling with proper height calculation
- Updated projects to start with a default "Default" layer for better user experience
- Fixed layer initialization: default layer is now created inline in the store and `clearLayers()` resets to an empty array, so loading a project replaces layers entirely while a fresh startup keeps the default
- Removed `clearProject()` from `useProject` cleanup to avoid wiping state on unmount

## Use of AI

AI was mainly used in the `AddDataObjectDialog` component (and for bug fixing)

### DIFF
--- a/apps/tissuumaps/src/App.tsx
+++ b/apps/tissuumaps/src/App.tsx
@@ -28,12 +28,24 @@ const dockviewTheme: DockviewTheme = {
 
 const dockviewComponents = {
   ViewerPanel: () => <ViewerPanel className="size-full" />,
-  ProjectPanel: () => <ProjectPanel className="m-2" />,
-  ImagesPanel: () => <ImagesPanel className="m-2" />,
-  LabelsPanel: () => <LabelsPanel className="m-2" />,
-  PointsPanel: () => <PointsPanel className="m-2" />,
-  ShapesPanel: () => <ShapesPanel className="m-2" />,
-  TablesPanel: () => <TablesPanel className="m-2" />,
+  ProjectPanel: () => (
+    <ProjectPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
+  ),
+  ImagesPanel: () => (
+    <ImagesPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
+  ),
+  LabelsPanel: () => (
+    <LabelsPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
+  ),
+  PointsPanel: () => (
+    <PointsPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
+  ),
+  ShapesPanel: () => (
+    <ShapesPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
+  ),
+  TablesPanel: () => (
+    <TablesPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
+  ),
 };
 
 const dockviewTabComponents = {
--- a/apps/tissuumaps/src/App.tsx
+++ b/apps/tissuumaps/src/App.tsx
@@ -28,24 +28,12 @@ const dockviewTheme: DockviewTheme = {
 
 const dockviewComponents = {
   ViewerPanel: () => <ViewerPanel className="size-full" />,
-  ProjectPanel: () => (
-    <ProjectPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
-  ),
-  ImagesPanel: () => (
-    <ImagesPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
-  ),
-  LabelsPanel: () => (
-    <LabelsPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
-  ),
-  PointsPanel: () => (
-    <PointsPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
-  ),
-  ShapesPanel: () => (
-    <ShapesPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
-  ),
-  TablesPanel: () => (
-    <TablesPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
-  ),
+  ProjectPanel: () => <ProjectPanel className="m-2" />,
+  ImagesPanel: () => <ImagesPanel className="m-2" />,
+  LabelsPanel: () => <LabelsPanel className="m-2" />,
+  PointsPanel: () => <PointsPanel className="m-2" />,
+  ShapesPanel: () => <ShapesPanel className="m-2" />,
+  TablesPanel: () => <TablesPanel className="m-2" />,
 };
 
 const dockviewTabComponents = {
--- a/apps/tissuumaps/src/components/common/simple-combobox.tsx
+++ b/apps/tissuumaps/src/components/common/simple-combobox.tsx
@@ -176,7 +176,7 @@ export function SimpleAsyncCombobox<TItem>({
       </div>
       <ComboboxPrimitive.Portal>
         <ComboboxPrimitive.Positioner
-          className="outline-none z-60"
+          className="outline-none isolate z-50"
           sideOffset={4}
         >
           <ComboboxPrimitive.Popup
--- a/apps/tissuumaps/src/components/common/simple-combobox.tsx
+++ b/apps/tissuumaps/src/components/common/simple-combobox.tsx
@@ -175,7 +175,10 @@ export function SimpleAsyncCombobox<TItem>({
         </div>
       </div>
       <ComboboxPrimitive.Portal>
-        <ComboboxPrimitive.Positioner className="outline-none" sideOffset={4}>
+        <ComboboxPrimitive.Positioner
+          className="outline-none z-60"
+          sideOffset={4}
+        >
           <ComboboxPrimitive.Popup
             className="box-border w-(--anchor-width) max-h-[min(var(--available-height),23rem)] max-w-(--available-width) origin-(--transform-origin) overflow-y-auto scroll-pb-2 scroll-pt-2 overscroll-contain rounded-md bg-[canvas] py-2 text-foreground shadow-[0_10px_15px_-3px_var(--color-gray-200),0_4px_6px_-4px_var(--color-gray-200)] outline-1 outline-gray-200 transition-[transform,scale,opacity] data-ending-style:transition-none data-starting-style:scale-95 data-starting-style:opacity-0 dark:bg-input/30 dark:-outline-offset-1 dark:shadow-none dark:outline-gray-300"
             aria-busy={isSearchPending || isSelectPending || undefined}
--- a/apps/tissuumaps/src/components/panels/ImagesPanel/index.tsx
+++ b/apps/tissuumaps/src/components/panels/ImagesPanel/index.tsx
@@ -18,7 +18,7 @@ import {
   InputGroupAddon,
   InputGroupInput,
 } from "@/components/ui/input-group";
-import { AddDataSourceDialog } from "@/components/widgets/AddDataSourceDialog";
+import { AddDataObjectDialog } from "@/components/widgets/AddDataObjectDialog";
 import { DataSourceWidget } from "@/components/widgets/DataSourceWidget";
 import { cn } from "@/lib/utils";
 import { useTissUUmaps } from "@/store";
@@ -55,14 +55,14 @@ export function ImagesPanel({ className }: ImagesPanelProps) {
           ))}
         </Accordion>
       </DragDropProvider>
-      <AddDataSourceDialog
+      <AddDataObjectDialog
         title="Add image"
         dataProviders={imageDataProviders}
         onAdd={(name, _type, dataSource) => {
           const image = createImage({
             id: crypto.randomUUID(),
             name,
-            dataSource: dataSource as Image["dataSource"],
+            dataSource,
             layerConfigs: layers.length > 0 ? [{ layer: layers[0]!.id }] : [],
           });
           addImage(image);
--- a/apps/tissuumaps/src/components/panels/ImagesPanel/index.tsx
+++ b/apps/tissuumaps/src/components/panels/ImagesPanel/index.tsx
@@ -34,7 +34,6 @@ export function ImagesPanel({ className }: ImagesPanelProps) {
   const layers = useTissUUmaps((state) => state.layers);
   const imageDataProviders = useTissUUmaps((state) => state.imageDataProviders);
   const addImage = useTissUUmaps((state) => state.addImage);
-  const loadImage = useTissUUmaps((state) => state.loadImage);
   const moveImage = useTissUUmaps((state) => state.moveImage);
 
   return (
@@ -66,7 +65,6 @@ export function ImagesPanel({ className }: ImagesPanelProps) {
             layerConfigs: layers.length > 0 ? [{ layer: layers[0]!.id }] : [],
           });
           addImage(image);
-          loadImage(image.id).catch(console.error);
         }}
       />
     </div>
--- a/apps/tissuumaps/src/components/panels/ImagesPanel/index.tsx
+++ b/apps/tissuumaps/src/components/panels/ImagesPanel/index.tsx
@@ -2,7 +2,7 @@ import { DragDropProvider } from "@dnd-kit/react";
 import { isSortable, useSortable } from "@dnd-kit/react/sortable";
 import { EyeIcon, EyeOffIcon, GripVertical, Trash2Icon } from "lucide-react";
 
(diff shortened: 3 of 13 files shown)


apps/tissuumaps/src/store/slices/project.ts
```diff
@@ -71,6 +71,7 @@ export const createProjectSlice: TissUUmapsStateCreator<ProjectSlice> = (
       get().clearProject();
       set({
         projectName: project.name,
+        layers: [],
         markerMaps: structuredClone(project.markerMaps),
```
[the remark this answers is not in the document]
cavenel: Correct projects should always have a saved layer. We only add a default layer when no project is loaded. Should be fine.

apps/tissuumaps/src/components/widgets/AddDataSourceDialog/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { cells } from "../DataSourceWidget/cells";
+import { renderers } from "../DataSourceWidget/renderers";
+
+export type AddDataSourceDialogProps = {
+  title: string;
+  dataProviders: Map<string, DataProvider<DataSource, Data>>;
+  onAdd: (name: string, dataSourceType: string, dataSource: DataSource) => void;
+};
+
+export function AddDataSourceDialog({
```
jwindhager: Since we're adding a data object (image/labels/points/shapes/table) here: Can we name it `AddDataObjectDialog` instead?

apps/tissuumaps/src/components/widgets/AddDataSourceDialog/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { Input } from "@/components/ui/input";
+import { Label } from "@/components/ui/label";
+import { RadioGroup, RadioGroupItem } from "@/components/ui/radio-group";
+
+import { cells } from "../DataSourceWidget/cells";
+import { renderers } from "../DataSourceWidget/renderers";
+
+export type AddDataSourceDialogProps = {
+  title: string;
+  dataProviders: Map<string, DataProvider<DataSource, Data>>;
```
jwindhager: I think this should be generically typed (at least for `DataSource`, perhaps also for `DataProvider` and `Data`), such that e.g. `onAdd["dataSource"]` can be of type `TDataSource`

apps/tissuumaps/src/components/panels/ImagesPanel/index.tsx
```diff
@@ -29,25 +31,45 @@ export type ImagesPanelProps = {
...
+        </Accordion>
+      </DragDropProvider>
+      <AddDataSourceDialog
+        title="Add image"
+        dataProviders={imageDataProviders}
+        onAdd={(name, _type, dataSource) => {
+          const image = createImage({
+            id: crypto.randomUUID(),
+            name,
+            dataSource: dataSource as Image["dataSource"],
```
jwindhager: Type cast should not be necessary when using generics

apps/tissuumaps/src/components/panels/ImagesPanel/index.tsx
```diff
@@ -29,25 +31,45 @@ export type ImagesPanelProps = {
...
+            layerConfigs: layers.length > 0 ? [{ layer: layers[0]!.id }] : [],
+          });
+          addImage(image);
+          loadImage(image.id).catch(console.error);
```
jwindhager: I don't think this is necessary - the rendering logic should take care of that

```suggestion
```

apps/tissuumaps/src/components/panels/LabelsPanel/index.tsx
```diff
@@ -29,27 +31,49 @@ export type LabelsPanelProps = {
...
+        </Accordion>
+      </DragDropProvider>
+      <AddDataSourceDialog
+        title="Add labels"
+        dataProviders={labelsDataProviders}
+        onAdd={(name, _type, dataSource) => {
+          const newLabels = createLabels({
+            id: crypto.randomUUID(),
+            name,
+            dataSource: dataSource as Labels["dataSource"],
```
jwindhager: Type cast should not be necessary when using generics

apps/tissuumaps/src/components/panels/PointsPanel/index.tsx
```diff
@@ -29,29 +31,52 @@ export type PointsPanelProps = {
...
+        </Accordion>
+      </DragDropProvider>
+      <AddDataSourceDialog
+        title="Add points"
+        dataProviders={pointsDataProviders}
+        onAdd={(name, _type, dataSource) => {
+          const newPoints = createPoints({
+            id: crypto.randomUUID(),
+            name,
+            dataSource: dataSource as Points["dataSource"],
```
jwindhager: Type cast should not be necessary when using generics

apps/tissuumaps/src/components/panels/ShapesPanel/index.tsx
```diff
@@ -29,29 +31,51 @@ export type ShapesPanelProps = {
...
+        </Accordion>
+      </DragDropProvider>
+      <AddDataSourceDialog
+        title="Add shapes"
+        dataProviders={shapesDataProviders}
+        onAdd={(name, _type, dataSource) => {
+          const newShapes = createShapes({
+            id: crypto.randomUUID(),
+            name,
+            dataSource: dataSource as Shapes["dataSource"],
```
jwindhager: Type cast should not be necessary when using generics

apps/tissuumaps/src/components/panels/TablesPanel/index.tsx
```diff
@@ -24,23 +26,41 @@ export type TablesPanelProps = {
...
+        </Accordion>
+      </DragDropProvider>
+      <AddDataSourceDialog
+        title="Add table"
+        dataProviders={tableDataProviders}
+        onAdd={(name, _type, dataSource) => {
+          const table = createTable({
+            id: crypto.randomUUID(),
+            name,
+            dataSource: dataSource as Table["dataSource"],
```
jwindhager: Type cast should not be necessary when using generics

apps/tissuumaps/src/components/panels/LabelsPanel/index.tsx
```diff
@@ -29,27 +31,49 @@ export type LabelsPanelProps = {
...
+            layerConfigs: layers.length > 0 ? [{ layer: layers[0]!.id }] : [],
+          });
+          addLabels(newLabels);
+          loadLabels(newLabels.id).catch(console.error);
```
jwindhager: I don't think this is necessary - the rendering logic should take care of that


```suggestion
```

apps/tissuumaps/src/components/panels/PointsPanel/index.tsx
```diff
@@ -29,29 +31,52 @@ export type PointsPanelProps = {
...
+            layerConfigs:
+              layers.length > 0 ? [{ layer: layers[0]!.id, x: "", y: "" }] : [],
+          });
+          addPoints(newPoints);
+          loadPoints(newPoints.id).catch(console.error);
```
jwindhager: I don't think this is necessary - the rendering logic should take care of that


```suggestion
```

apps/tissuumaps/src/components/panels/ShapesPanel/index.tsx
```diff
@@ -29,29 +31,51 @@ export type ShapesPanelProps = {
...
+            layerConfigs: layers.length > 0 ? [{ layer: layers[0]!.id }] : [],
+          });
+          addShapes(newShapes);
+          loadShapes(newShapes.id).catch(console.error);
```
jwindhager: I don't think this is necessary - the rendering logic should take care of that

```suggestion
```

apps/tissuumaps/src/components/panels/TablesPanel/index.tsx
```diff
@@ -24,23 +26,41 @@ export type TablesPanelProps = {
...
+          });
+          addTable(table);
+          loadTable(table.id).catch(console.error);
```
jwindhager: I don't think this is necessary - the rendering logic should take care of that

```suggestion
```

apps/tissuumaps/src/components/widgets/AddDataObjectDialog/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+  dataProviders: Map<string, DataProvider<DataSource, Data>>;
+  onAdd: (name: string, dataSourceType: string, dataSource: DataSource) => void;
+};
+
+export function AddDataSourceDialog({
+  title,
+  dataProviders,
+  onAdd,
+}: AddDataSourceDialogProps) {
+  const providerEntries = Array.from(dataProviders.entries());
```
jwindhager, line 46: Could we get rid of `providerEntries` and directly use `dataProviders.keys()`, `dataProviders.values()` and `dataProviders.entries()` where needed? I think this would improve readability

apps/tissuumaps/src/components/widgets/AddDataObjectDialog/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const handleTypeChange = useCallback((value: string) => {
+    setSelectedType(value);
+    setDataSourceDraft({ type: value });
+  }, []);
+
+  const handleAdd = useCallback(() => {
+    onAdd(name.trim() || "Untitled", selectedType, dataSourceDraft);
+    setOpen(false);
+    resetForm();
+  }, [name, selectedType, dataSourceDraft, onAdd, resetForm]);
```
jwindhager, line 74: Why wrap these three in `useCallback`? Can't we directly define them as "anonymous" event handlers within the JSX code?

apps/tissuumaps/src/components/widgets/AddDataSourceDialog/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+                data={dataSourceDraft}
+                onChange={({ data }) => setDataSourceDraft(data as DataSource)}
+                schema={selectedProvider.schema}
+                uischema={selectedProvider.uischema}
+                renderers={renderers}
+                cells={cells}
+              />
+            </div>
+          )}
+        </div>
```
jwindhager: This section (all the form components) to me looks like a potentially good use case for `Fieldset` (outer div) with `Field` (inner div), `FieldLabel` (Label) and other input components

apps/tissuumaps/src/components/widgets/AddDataSourceDialog/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+                {providerEntries.map(([type, provider]) => {
+                  const radioId = `data-source-type-${type}`;
+                  return (
+                    <div key={type} className="flex items-center gap-2">
+                      <RadioGroupItem id={radioId} value={type} />
+                      <Label htmlFor={radioId}>{provider.name}</Label>
+                    </div>
+                  );
+                })}
+              </RadioGroup>
```
jwindhager: Do we want a `RadioGroup` or a `SimpleSelect` here?

apps/tissuumaps/src/store/slices/layer.ts
```diff
@@ -79,5 +79,12 @@ export const createLayerSlice: TissUUmapsStateCreator<LayerSlice> = (
...
 function createInitialLayerSliceState(): LayerSliceState {
-  return { layers: [] };
+  return {
+    layers: [
+      createLayer({
+        id: crypto.randomUUID(),
+        name: "Default",
+      }),
+    ],
+  };
```
jwindhager, line 89: Setting layers here will also add a default layer when clearing layers/projects. Perhaps the default layer should only be added at application start? In this case, this code would need to be moved directly into `createLayerSlice`

apps/tissuumaps/src/store/slices/project.ts
```diff
@@ -71,6 +71,7 @@ export const createProjectSlice: TissUUmapsStateCreator<ProjectSlice> = (
       get().clearProject();
       set({
         projectName: project.name,
+        layers: [],
```
jwindhager: When moving the default layer directly into createLayersSlice (see above), this will not be needed anymore I think

apps/tissuumaps/src/App.tsx
```diff
@@ -28,12 +28,24 @@ const dockviewTheme: DockviewTheme = {
...
+  ),
+  PointsPanel: () => (
+    <PointsPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
+  ),
+  ShapesPanel: () => (
+    <ShapesPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
+  ),
+  TablesPanel: () => (
+    <TablesPanel className="m-2 overflow-y-auto h-[calc(100%-1rem)]" />
+  ),
```
jwindhager: Could we perhaps move these changes into a separate PR?

jwindhager (changes requested):
Looks mostly good to me, just the AddDataSourceDialog itself could use another round of TLC I think 😇 

A few more questions:
- Why do we need to set `z-index` in SimpleSelect/SimpleCombobox?
- Could we create the id based on the user-provided name (sanitized) instead of UUID?
- UI: I think it would be nice if the button would be sticky in the bottom part of the panels

apps/tissuumaps/src/components/widgets/AddDataSourceDialog/index.tsx
cavenel: Yes, done!

apps/tissuumaps/src/components/widgets/AddDataSourceDialog/index.tsx
cavenel: Yes, done!

apps/tissuumaps/src/components/widgets/AddDataSourceDialog/index.tsx
cavenel: Yes, done!

cavenel:
> * Why do we need to set `z-index` in SimpleSelect/SimpleCombobox?

When the SimpleSelect dropdown opens inside a modal Dialog, both the dialog and the dropdown create portals, they inject their HTML into the document body. The dialog's overlay uses z-50 to sit on top of everything. The dropdown portal ends up as a sibling or nested element that competes with that same z-50 layer. Without a higher z-index (like z-60), the dropdown would appear behind the dialog overlay, making it invisible or unclickable.

We could fix by making sure that the dropdown is created *inside* the container of the dialog. That would be a bit of changes though.

For now, we align with shadcn conventions and use "z-50 isolate"

> * Could we create the id based on the user-provided name (sanitized) instead of UUID?

We could. Would it be better?

> * UI: I think it would be nice if the button would be sticky in the bottom part of the panels

I think I would like to keep that for when we add the scrollbar in the right menu then.

apps/tissuumaps/src/store/slices/layer.ts
cavenel: Done!

apps/tissuumaps/src/components/common/simple-combobox.tsx
```diff
@@ -175,7 +175,10 @@ export function SimpleAsyncCombobox<TItem>({
         </div>
       </div>
       <ComboboxPrimitive.Portal>
-        <ComboboxPrimitive.Positioner className="outline-none" sideOffset={4}>
+        <ComboboxPrimitive.Positioner
+          className="outline-none isolate z-50"
+          sideOffset={4}
+        >
```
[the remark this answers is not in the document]
cavenel: We assume it should be on top if loaded after the modal, and under if loaded before. I think this is fine.

apps/tissuumaps/src/components/panels/PointsPanel/index.tsx
```diff
@@ -33,29 +35,50 @@ export type PointsPanelProps = {
...
+        title="Add points"
+        dataProviders={pointsDataProviders}
+        onAdd={(name, _type, dataSource) => {
+          const newPoints = createPoints({
+            id: crypto.randomUUID(),
+            name,
+            dataSource,
+            layerConfigs:
+              layers.length > 0 ? [{ layer: layers[0]!.id, x: "", y: "" }] : [],
+          });
```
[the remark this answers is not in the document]
cavenel: We will default to "x" and "y" for now

jwindhager (approved):
LGTM

Add gap in LabelsPanel for consistency
Update description to mention the default layer

PR #113, opened 2026-04-08
feat: implement rectangle shape drawing

HenrikeW:

Closes [TMAP-221](https://scilifelab.atlassian.net/browse/TMAP-221)

# Type of change

- [ ] Other

# List of changes made

- added mouse event handling to the SVGController to draw a rectangle
- modified OpenSeadragon mouse navigation in draw mode (panning now done by using Shift + pointerdown

# Screenshot of the fix

image

# Use of AI
Claude Opus 4.5 was used during development for guidance and code suggestions. The suggestions were tested and modified manually.

# Testing
- switch to draw rectangle mode
- make sure you can draw one rectangle at a time (it should disappear when you stop pressing the mouse button)
- load a project
- make sure you can draw one rectangle at a time on it
- you should be able to start and finish the rectangle anywhere in the viewing area. 
- you should still be able to move your project image by pressing shift and dragging at the same time
- when zooming in, the border of the rectangle should keep it's absolute width (not become thicker)

# Further comments

### DIFF
--- a/packages/@tissuumaps-core/src/controllers/SVGController.ts
+++ b/packages/@tissuumaps-core/src/controllers/SVGController.ts
@@ -7,11 +7,11 @@ import {
 
 const SVG_NAMESPACE = "http://www.w3.org/2000/svg";
 
-interface DrawingState {
+type DrawingState = {
   isDrawing: boolean;
   startPoint: Vertex | null;
   currentRect: SVGRectElement | null;
-}
+};
 
 /**
  * Controller for managing the drawing of SVG shapes
@@ -20,6 +20,10 @@ export class SVGController {
   public readonly container: SVGSVGElement;
   public readonly transformNode: SVGGElement;
   public readonly shapeCompleteHandler?: (shape: MultiPolygon) => void;
+  private static readonly _previewFillColor = "rgba(255, 0, 0, 0.2)";
+  private static readonly _previewStrokeColor = "#FF0000";
+  private static readonly _previewStrokeWidthFactor = 0.0005;
+  private static readonly _minShapeSizeFactor = 0.001;
   private _containerSize: { width: number; height: number };
   private _viewport: Rect;
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
@@ -167,13 +171,17 @@ export class SVGController {
   };
 
   private _handlePointerMove = (event: PointerEvent): void => {
-    if (!this._drawingState.isDrawing || !this._drawingState.currentRect)
+    if (
+      !this._drawingState.isDrawing ||
+      !this._drawingState.currentRect ||
+      !this._drawingState.startPoint
+    )
       return;
 
     const worldPoint = this._screenToWorld(event.clientX, event.clientY);
     this._updatePreviewRect(
       this._drawingState.currentRect,
-      this._drawingState.startPoint!,
+      this._drawingState.startPoint,
       worldPoint,
     );
   };
@@ -194,7 +202,16 @@ export class SVGController {
       }
     }
 
-    this._finishDrawing();
+    // Remove preview element - WebGL will render the actual shape
+    this._drawingState.currentRect?.remove();
+    this._drawingState = {
+      isDrawing: false,
+      startPoint: null,
+      currentRect: null,
+    };
+
+    document.removeEventListener("pointermove", this._handlePointerMove);
+    document.removeEventListener("pointerup", this._handlePointerUp);
   };
 
   // ─────────────────────────────────────────────────────────────
@@ -207,13 +224,14 @@ export class SVGController {
     rect.setAttribute("y", startPoint.y.toString());
     rect.setAttribute("width", "0");
     rect.setAttribute("height", "0");
-    rect.setAttribute("fill", "rgba(255, 0, 0, 0.2)");
-    rect.setAttribute("stroke", "#FF0000");
+    rect.setAttribute("fill", SVGController._previewFillColor);
+    rect.setAttribute("stroke", SVGController._previewStrokeColor);
     rect.setAttribute(
       "stroke-width",
-      (this._viewport.width * 0.0005).toString(),
+      (
+        this._viewport.width * SVGController._previewStrokeWidthFactor
+      ).toString(),
     );
-    rect.classList.add("drawing-preview");
 
     this.transformNode.appendChild(rect);
     return rect;
@@ -239,19 +257,19 @@ export class SVGController {
     start: Vertex,
     end: Vertex,
   ): MultiPolygon {
-    const x1 = Math.min(start.x, end.x);
-    const y1 = Math.min(start.y, end.y);
-    const x2 = Math.max(start.x, end.x);
-    const y2 = Math.max(start.y, end.y);
+    const x0 = Math.min(start.x, end.x);
+    const y0 = Math.min(start.y, end.y);
+    const x1 = Math.max(start.x, end.x);
+    const y1 = Math.max(start.y, end.y);
 
     return {
       polygons: [
         {
           shell: [
+            { x: x0, y: y0 },
+            { x: x1, y: y0 },
             { x: x1, y: y1 },
-            { x: x2, y: y1 },
-            { x: x2, y: y2 },
-            { x: x1, y: y2 },
+            { x: x0, y: y1 },
           ],
           holes: [],
         },
@@ -260,23 +278,10 @@ export class SVGController {
   }
 
   private _hasMinimumSize(start: Vertex, end: Vertex): boolean {
-    const minSize = this._viewport.width * 0.001; // Adjust threshold as needed
+    const minSize = this._viewport.width * SVGController._minShapeSizeFactor; // Adjust threshold as needed
     return (
       Math.abs(end.x - start.x) > minSize && Math.abs(end.y - start.y) > minSize
     );
-  }
-
-  private _finishDrawing(): void {
-    // Remove preview element - WebGL will render the actual shape
-    this._drawingState.currentRect?.remove();
-    this._drawingState = {
-      isDrawing: false,
-      startPoint: null,
-      currentRect: null,
-    };
-
-    document.removeEventListener("pointermove", this._handlePointerMove);
-    document.removeEventListener("pointerup", this._handlePointerUp);
   }
 
   // ─────────────────────────────────────────────────────────────
--- a/packages/@tissuumaps-core/src/controllers/SVGController.ts
+++ b/packages/@tissuumaps-core/src/controllers/SVGController.ts
@@ -11,9 +11,7 @@ export class SVGController {
   public readonly shapeCompleteHandler?: (shape: MultiPolygon) => void;
   private _containerSize: { width: number; height: number };
   private _viewport: Rect;
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore currently not used, but will be needed in the future
-  private _interactionMode?: InteractionMode;
+  private _interactionMode: InteractionMode = "pan";
 
   /** Creates a positioned, full-size `<svg>` element for the SVG overlay */
   static createContainer(): SVGSVGElement {
@@ -23,6 +21,7 @@ export class SVGController {
     container.style.left = "0";
     container.style.width = "100%";
     container.style.height = "100%";
+    container.style.pointerEvents = "none";
     return container;
   }
 
@@ -45,7 +44,6 @@ export class SVGController {
     };
     this._viewport = initialViewport;
     container.replaceChildren(this.transformNode);
-    // TODO register mouse event handlers on container
   }
 
   /**
@@ -69,12 +67,17 @@ export class SVGController {
   }
 
   /**
-   * Updates the interaction mode
+   * Updates the interaction mode. In "pan" mode, pointer events pass through
+   * to the underlying OpenSeadragon canvas. In drawing modes, the SVG overlay
+   * captures pointer events to handle shape drawing.
    *
    * @param newInteractionMode - New interaction mode
    */
   setInteractionMode(newInteractionMode: InteractionMode) {
+    if (this._interactionMode === newInteractionMode) return;
     this._interactionMode = newInteractionMode;
+    this.container.style.pointerEvents =
+      newInteractionMode === "pan" ? "none" : "auto";
   }
 
   /**
@@ -108,10 +111,9 @@ export class SVGController {
     return false;
   }
 
-  destroy(): void {}
-
-  // TODO implement mouse event handlers for shape drawing; upon shape completion, call shapeCompleteHandler (if defined);
-  // mouse coordinates can be transformed to world space coordinates using transformNode.getScreenCTM().inverse()
+  destroy(): void {
+    this.setInteractionMode("pan");
+  }
 
   /**
    * Updates the world-to-viewport transform based on the current container size
--- a/packages/@tissuumaps-core/src/controllers/SVGController.ts
+++ b/packages/@tissuumaps-core/src/controllers/SVGController.ts
@@ -136,25 +136,23 @@ export class SVGController {
   // ─────────────────────────────────────────────────────────────
 
   private _registerEventHandlers(): void {
-    console.debug("Registering SVG event handlers");
     this.container.addEventListener("pointerdown", this._handlePointerDown);
-    document.addEventListener("pointermove", this._handlePointerMove);
-    document.addEventListener("pointerup", this._handlePointerUp);
   }
 
   private _unregisterEventHandlers(): void {
     this.container.removeEventListener("pointerdown", this._handlePointerDown);
-    document.removeEventListener("pointermove", this._handlePointerMove);
-    document.removeEventListener("pointerup", this._handlePointerUp);
   }
 
   private _handlePointerDown = (event: PointerEvent): void => {
-    if (this._interactionMode !== "draw") return;
+    if (this._interactionMode !== "drawRectangle") return;
     if (event.button !== 0) return;
     if (event.shiftKey) {
       // Shift held = don't draw, let instead OpenSeadragon handle panning
       return;
     }
+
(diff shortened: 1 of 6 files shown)


cavenel:
> * Currently, the whole mouse navigation of OpenSeadragon is switched off in draw mode, which means no zooming, no moving the image etc.
>   
>   * Is this acceptable and feasible in the final UI?
>   * Any better ways to prevent the mouse event from going through to the OpenSeadragon canvas?
>   * Should at least some mouse events stay, like scrolling for zooming?

In TissUUmaps 3, only mouse dragging was disabled, allowing user to move around by scrolling to zoom/unzoom. I think it would be nice to have the same behavior, to not be stuck on one viewport when drawing the rectangle.

> * Should anything happen on mouse leave?

It would be nice that it continues to move the rectangle, even outside the svg? See https://tissuumaps-demo.serve.scilifelab.se/human_heart.tmap for example. You could attach the mousemove and mouseup to the document instead of the container only: `document.addEventListener("mousemove", this._handleMouseMove);` etc.

> * Any requirements for the color of the rectangle?

Not really. Maybe transparent, and thick red edges?

cavenel:
To block only the drag and drop, but keep zoom, I think you can use:
```javascript
this.viewer.gestureSettingsByDeviceType("mouse").dragToPan = enabled;
this.viewer.gestureSettingsByDeviceType("touch").dragToPan = enabled;
this.viewer.gestureSettingsByDeviceType("pen").dragToPan = enabled;
this.viewer.gestureSettingsByDeviceType("unknown").dragToPan = enabled;
```
and use `PointerEvent` (with `pointerup`, `pointerdown`, etc.) so that the events still work correctly.

HenrikeW:
Okay, I'll look into only disabling the dragging then, that sounds reasonable I think.

About continuing the drawing of rectangles outside of the SVG container, I'm a bit more hesitant. Honestly, I think in TissUUmaps 3 it's a weird user experience seeing cut-off rectangles like this:
image
Is there any case where they are useful for the user, or even needed? Maybe I know too little about how the site is actually being used. Otherwise I think it would make sense to limit the shapes to where they actually can be displayed.

cavenel:
In theory, with the new version, shapes can be drawn everywhere. So we should not see these artifcts anymore!
I think by experience, that it is frustrating for users when things get stuck as soon as the mouse leave momentarily the viewport.

HenrikeW:
I see, thanks for explaining. I implemented both of your suggestions now. 
I realized that there is an edge case in the drawing: When the users happen to start their rectangle on top of the navigation box in the lower left corner, the pointer events still go through and they are moving the image instead of drawing a rectangle. This can happen quite easily as the box fades out after a couple of seconds and becomes invisible. Do you want me to look into that edge case now or is that  fix that you want to leave for later?

Also, if we would want to make even dragging possible in draw mode, could it be an option to do it "the Miro way" and enable dragging by right clicking? It's an idea that struck my mind while working on it but I'm not sure if it breaks too much with users' habits on TissUUmaps.

cavenel:
> When the users happen to start their rectangle on top of the navigation box in the lower left corner, the pointer events still go through and they are moving the image instead of drawing a rectangle. This can happen quite easily as the box fades out after a couple of seconds and becomes invisible. Do you want me to look into that edge case now or is that fix that you want to leave for later?

I think this is an ok behavior. The navigator should come back as soon as the user moves the mouse on top of the viewer. So the issue will only happen for people using touch screen I guess?
Easy fix would be to make the navigator always visible, but I don't think it is needed. Touch screen users that want to draw are probably not the majority of users.

> Also, if we would want to make even dragging possible in draw mode, could it be an option to do it "the Miro way" and enable dragging by right clicking? It's an idea that struck my mind while working on it but I'm not sure if it breaks too much with users' habits on TissUUmaps.

That is a good idea, could be right click or shift key pressed?
I think it is ok to change users' habits, if it offers more features and intuitive ways of working! :)

HenrikeW:
Cool! The right-click thing turned out to be quite complicated but drag with shift key pressed was an easy fix, so I went for that. I think a short information message about that option should pop up somewhere in the UI when the user is changing into draw mode but I'd consider that outside of the scope of this PR.

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +126,177 @@ export class SVGController {
...
+
+  private _createPreviewRect(startPoint: Vertex): SVGRectElement {
+    const rect = document.createElementNS(SVG_NAMESPACE, "rect");
+    rect.setAttribute("x", startPoint.x.toString());
+    rect.setAttribute("y", startPoint.y.toString());
+    rect.setAttribute("width", "0");
+    rect.setAttribute("height", "0");
+    rect.setAttribute("fill", "rgba(255, 0, 0, 0.2)");
+    rect.setAttribute("stroke", "#FF0000");
+    rect.setAttribute("stroke-width", "20");
```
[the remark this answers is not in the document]
jwindhager: @cavenel I think the concern regarding different zoom levels is valid; how is this done in TM3?

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -45,7 +62,8 @@ export class SVGController {
     };
     this._viewport = viewport;
     container.replaceChildren(this.transformNode);
-    // TODO register mouse event handlers on container
+
+    this._registerEventHandlers();
   }
```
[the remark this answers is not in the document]
jwindhager: https://github.com/TissUUmaps/TissUUmaps4/pull/118

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +126,177 @@ export class SVGController {
...
+
+  // ─────────────────────────────────────────────────────────────
+  // Coordinate Transformation
+  // ─────────────────────────────────────────────────────────────
+
+  private _screenToWorld(screenX: number, screenY: number): Vertex {
+    const ctm = this.transformNode.getScreenCTM();
+    if (!ctm) {
+      console.warn("Could not get screen CTM");
+      return { x: 0, y: 0 };
```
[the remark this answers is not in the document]
jwindhager: Now that https://github.com/TissUUmaps/TissUUmaps4/pull/118 is merged, I think this should indeed throw an error rather than failing silently

packages/@tissuumaps-core/src/controllers/SVGController.ts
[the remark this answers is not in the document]
HenrikeW: I think this is a valid point. I'll replace it with a width that adjusts to the zoom level. 
```suggestion
    rect.setAttribute("fill", "rgba(255, 0, 0, 0.2)");
    rect.setAttribute("stroke", "#FF0000");
        rect.setAttribute(
      "stroke-width",
      (this._viewport.width * 0.0005).toString(),
    );
```

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +126,177 @@ export class SVGController {
...
+  // ─────────────────────────────────────────────────────────────
+  // Event Handling
+  // ─────────────────────────────────────────────────────────────
+
+  private _registerEventHandlers(): void {
+    console.debug("Registering SVG event handlers");
+    this.container.addEventListener("pointerdown", this._handlePointerDown);
+    document.addEventListener("pointermove", this._handlePointerMove);
+    document.addEventListener("pointerup", this._handlePointerUp);
+  }
```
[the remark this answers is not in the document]
HenrikeW: I thought this was a good suggestion so I made a change according to it.

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +126,177 @@ export class SVGController {
...
+      isDrawing: true,
+      startPoint: worldPoint,
+      currentRect: this._createPreviewRect(worldPoint),
+    };
+  };
+
+  private _handlePointerMove = (event: PointerEvent): void => {
+    if (!this._drawingState.isDrawing || !this._drawingState.currentRect)
+      return;
+
```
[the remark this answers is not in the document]
HenrikeW: I don't think this would be a scenario that we need to take into account and would like to ignore this comment. Please let me know if you think I should address it anyway.

packages/@tissuumaps-core/src/controllers/SVGController.ts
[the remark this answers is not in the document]
cavenel: Yes, TM3 takes the zoom value into account, see https://github.com/TissUUmaps/TissUUmaps/blob/618b97caf0767bae5b2fc8580c9a29e203607fd8/tissuumaps/static/js/utils/regionOperations.js#L45
So something like:
`const strokeWstr = polygonStrokeWidth / this._viewport.getZoom();`
I don't know if that would be the same as using the viewport width.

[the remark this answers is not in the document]
HenrikeW: According to Claude, both will achieve the same visual result :)

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -1,7 +1,18 @@
...
+  type Vertex,
+} from "../types";
 
 const SVG_NAMESPACE = "http://www.w3.org/2000/svg";
 
+interface DrawingState {
+  isDrawing: boolean;
+  startPoint: Vertex | null;
+  currentRect: SVGRectElement | null;
+}
```
jwindhager: I agree that we probably should've more leaned to the `interface` side, but so far we mostly use `type` across the codebase. For consistency, could we do:
```suggestion
type DrawingState = {
  isDrawing: boolean;
  startPoint: Vertex | null;
  currentRect: SVGRectElement | null;
};
```

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +124,180 @@ export class SVGController {
...
+    this.container.addEventListener("pointerdown", this._handlePointerDown);
+  }
+
+  private _unregisterEventHandlers(): void {
+    this.container.removeEventListener("pointerdown", this._handlePointerDown);
+  }
+
+  private _handlePointerDown = (event: PointerEvent): void => {
+    if (this._interactionMode !== "drawRectangle") return;
+    if (event.button !== 0) return;
```
jwindhager, line 152: What does this do?

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +124,180 @@ export class SVGController {
...
+  };
+
+  private _handlePointerMove = (event: PointerEvent): void => {
+    if (!this._drawingState.isDrawing || !this._drawingState.currentRect)
+      return;
+
+    const worldPoint = this._screenToWorld(event.clientX, event.clientY);
+    this._updatePreviewRect(
+      this._drawingState.currentRect,
+      this._drawingState.startPoint!,
```
jwindhager: Should we add an additional guard to the initial `if` checking for `startPoint` instead of the non-null assertion here?

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +124,180 @@ export class SVGController {
...
+    rect.setAttribute("y", startPoint.y.toString());
+    rect.setAttribute("width", "0");
+    rect.setAttribute("height", "0");
+    rect.setAttribute("fill", "rgba(255, 0, 0, 0.2)");
+    rect.setAttribute("stroke", "#FF0000");
+    rect.setAttribute(
+      "stroke-width",
+      (this._viewport.width * 0.0005).toString(),
+    );
+    rect.classList.add("drawing-preview");
```
jwindhager: Is this used anywhere?

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +124,180 @@ export class SVGController {
...
+    rect.setAttribute("x", startPoint.x.toString());
+    rect.setAttribute("y", startPoint.y.toString());
+    rect.setAttribute("width", "0");
+    rect.setAttribute("height", "0");
+    rect.setAttribute("fill", "rgba(255, 0, 0, 0.2)");
+    rect.setAttribute("stroke", "#FF0000");
+    rect.setAttribute(
+      "stroke-width",
+      (this._viewport.width * 0.0005).toString(),
+    );
```
jwindhager: Would be good to define fill/stroke/width factor as `static readonly` attributes of `SVGController`

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +124,180 @@ export class SVGController {
...
+  }
+
+  private _createRectangleMultiPolygon(
+    start: Vertex,
+    end: Vertex,
+  ): MultiPolygon {
+    const x1 = Math.min(start.x, end.x);
+    const y1 = Math.min(start.y, end.y);
+    const x2 = Math.max(start.x, end.x);
+    const y2 = Math.max(start.y, end.y);
```
jwindhager: Conventionally, we use zero-based point names (x0, y0, x1, y1), might be good to adopt this here as well for consistency (e.g. with the WebGL shaders)

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +124,180 @@ export class SVGController {
...
+            { x: x1, y: y2 },
+          ],
+          holes: [],
+        },
+      ],
+    };
+  }
+
+  private _hasMinimumSize(start: Vertex, end: Vertex): boolean {
+    const minSize = this._viewport.width * 0.001; // Adjust threshold as needed
```
jwindhager: Would be good to define this factor as `static readonly` attribute of `SVGController`

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -108,11 +124,180 @@ export class SVGController {
...
+    return (
+      Math.abs(end.x - start.x) > minSize && Math.abs(end.y - start.y) > minSize
+    );
+  }
+
+  private _finishDrawing(): void {
```
jwindhager: Currently, it's a bit unclear to me what is done in the pointer event handlers and what is done in separate functions called by them, but maybe that's just me 🙈. In this specific case, for example: Why not do this directly in the pointerup event handler?

jwindhager (changes requested):
Thanks a lot @HenrikeW, looks mostly good to me!

Only some minor "cosmetic" changes/clarifications from my side, see comments.

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: That's reasonable.

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: Line 147 makes sure nothing happens if you're not in draw rectangle mode.
Line 148 makes sure that you can only use the primary mouse button (left usually) for drawing. If you don't think it's necessary, I can remove it.

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: Good catch. Yes, we should. I changed it now.

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: No, this was left from an earlier stage. Good catch, I removed it now.

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: Yes that's a good idea I think. Changed it now.

packages/@tissuumaps-core/src/controllers/SVGController.ts
jwindhager: Ah, sorry, I was only referring to line 148. All good, was just wondering what the 0 meant :)

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: Yes definitely. I wasn't aware of that, thanks for pointing it out.

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: Fixed now.

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: My idea was to not put too much logic into the event handlers and most parts related to the drawing outside of it in helper functions for better readability. 

But in this case I think you're right. As the event listeners are added in the pointerDown event handler, it makes sense to also remove them inside the event handler itself. And a helper function only for the little cleanup removing the preview rectangle is maybe a bit overkill, so I moved it to the event handler.

But in general I would say this is mostly a question of taste, at least as long as the helper functions aren't reused. I guess we could put everything directly into the event handlers as well.

Maybe it becomes clearer what to prefer when we implement the drawing of other shades and free-hand.

jwindhager: Yes, that was my thinking - once we add more modes, it will become clearer what should be its own thing :)

I mostly stumbled upon this because only `_handlePointerUp()` seems to "outsource" core parts of the event handling, whereas the other event handlers do most "high-level" stuff inline.

You decide what's the best way forward here - also happy to keep it as is! As you say, it will probably change anyway...

jwindhager (approved):
Thanks for the changes, LGTM now!

PR #126, opened 2026-04-17
Prepare SVGController for polygon and free hand drawing

HenrikeW:

This PR is a preparation for TMAP-222 and TMAP-227

# Type of change

- [x] Other

# List of changes made

- made existing drawing functionality in SVGController specific for rectangle
- added placeholder functions and types for polygon and free hand drawing

# Use of AI

Claude Opus 4.5 was used as a chat bot during development.

# Testing

- Make sure that nothing broke.

# Further comments

This is a preparation so that we will be able to work in parallel on polygon drawing and free hand drawing. The idea is to avoid future merge conflicts and make our life easier.

### DIFF
--- a/packages/@tissuumaps-core/src/controllers/SVGController.ts
+++ b/packages/@tissuumaps-core/src/controllers/SVGController.ts
@@ -8,18 +8,15 @@ import {
 const SVG_NAMESPACE = "http://www.w3.org/2000/svg";
 
 type RectangleDrawingState = {
-  isDrawing: boolean;
-  startPoint: Vertex | null;
-  currentRect: SVGRectElement | null;
+  startPoint: Vertex;
+  currentRect: SVGRectElement;
 };
 
 type PolygonDrawingState = {
-  isDrawing: boolean;
   // TODO extend for polygon drawing
 };
 
 type FreehandDrawingState = {
-  isDrawing: boolean;
   // TODO extend for freehand drawing
 };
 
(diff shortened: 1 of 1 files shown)


jwindhager (commented):
Looking good!

The only feedback I have is that I think the `isDrawing` is redundant, expecially now that we have separate `DrawingState` types. One could for example infer `isDrawing` from whether `startPoint` or `currentRect` is not null (in the case of `RectangleDrawingState`).

To stick with the rectangle example (but the same applies to the other attributes): Even cleaner would probably be to make the `_rectangleDrawingState` attribute nullable/undefined (null/undefined --> not drawing, not null/defined --> drawing). In this case, the current `startPoint` and `currentRect` members would actually not need to be nullable anymore.

jwindhager (approved):
LGTM

Just one question: Can one have multiple drawing states being not null at the same time? If not, one could alternatively consider using just one attribute (union type?) shared by all drawing modes.

But I'd say good as it is for now, can always change it later if needed. Thanks for the PR!

HenrikeW:
> LGTM
> 
> Just one question: Can one have multiple drawing states being not null at the same time? If not, one could alternatively consider using just one attribute (union type?) shared by all drawing modes.
> 
> But I'd say good as it is for now, can always change it later if needed. Thanks for the PR!

Good question, I hadn't thought about that so far. For the rectangle I think it makes sense that it disappears (and the drawing state becomes null) as soon as you leave the mouse button. The polygon and free-hand shapes I suppose _could_ persist unfinished when the user switches drawing mode so that they could go back to it later but I don't know if that's a real use case.
So I think there's a good chance that only one drawing state at a time can be not-null and your suggestion then makes sense. I agree, that could be changed later, so I'll merge this PR for now.

PR #127, opened 2026-04-22
Point based shape drawing

HenrikeW:

This PR closes [TMAP-227](https://scilifelab.atlassian.net/browse/TMAP-227)

# Type of change

- [ ] Other

# List of changes made

- functionality for point based drawing was added
- shape will be drawn as polyline
- points are marked as small circles
- visual effect on starting point when coming close enough to close the shape
- keyboard listeners to finish and abort a shape drawing using keys

# Screenshot of the fix

image

# Use of AI
 Claude Opus 4.5 was used as a chat bot during development

# Testing

- switch to polygon mode
- start drawing with the mouse
- try to complete the polygon using the mouse
- try to complete the polygon using the keyboard
- try to abort the drawing using the keyboard
- does the behavior feel intuitive?

# Further comments

This PR is based on https://github.com/TissUUmaps/TissUUmaps4/pull/126 and should be merged afterwards.

### DIFF
--- a/packages/@tissuumaps-core/src/controllers/SVGController.ts
+++ b/packages/@tissuumaps-core/src/controllers/SVGController.ts
@@ -258,6 +258,10 @@ export class SVGController {
     if (!this._polygonDrawingState) {
       document.addEventListener("pointermove", this._handlePolygonPointerMove);
       document.addEventListener("keydown", this._handlePolygonKeyDown);
+      document.addEventListener(
+        "pointercancel",
+        this._handlePolygonPointerCancel,
+      );
 
       this._polygonDrawingState = {
         vertices: [worldPoint],
@@ -280,7 +284,7 @@ export class SVGController {
         this._createVertexMarker(worldPoint),
       );
 
-      this._updatePreviewPolyline(this._polygonDrawingState.vertices);
+      this._updatePolygonPreviewPolyline(this._polygonDrawingState.vertices);
     }
   };
 
@@ -299,7 +303,7 @@ export class SVGController {
       this._updateFirstVertexMarker(isNearStart);
     }
 
-    this._updatePreviewPolyline([
+    this._updatePolygonPreviewPolyline([
       ...this._polygonDrawingState.vertices,
       worldPoint,
     ]);
@@ -318,6 +322,11 @@ export class SVGController {
     } else if (this._polygonDrawingState.vertices.length >= 3) {
       this._completePolygon();
     }
+  };
+
+  private _handlePolygonPointerCancel = (): void => {
+    if (!this._polygonDrawingState) return;
+    this._cancelPolygon();
   };
 
   // ─────────────────────────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


cavenel:
Except for one minor comment, I think this can be merged!

HenrikeW:
> Except for one minor comment, I think this can be merged!

Great! Then I'll wait for PR #132 to be merged until I touch this one again.

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -244,25 +249,75 @@ export class SVGController {
...
+      document.addEventListener("keydown", this._handlePolygonKeyDown);
+
+      this._polygonDrawingState = {
+        vertices: [worldPoint],
+        currentPolyline: this._createPreviewPolyline(worldPoint),
+        vertexMarkers: [this._createVertexMarker(worldPoint)],
+        isNearStart: false,
+      };
+    } else {
+      const firstVertex = this._polygonDrawingState.vertices[0];
```
cavenel: We could move the `!` to the variable declaration, for consistency with the freehand drawing?

packages/@tissuumaps-core/src/controllers/SVGController.ts
cavenel, line 351: Should we also deal with `pointercancel` for polygon drawing?

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -532,6 +587,53 @@ export class SVGController {
     return dx * dx + dy * dy < threshold * threshold;
   }
 
+  // ─────────────────────────────────────────────────────────────
```
cavenel: All the freehand helpers are below the freehand handlers, for consistency we could do the same and move these polygon helpers under the polygon handlers? And maybe use naming that include "Polygon":`_updatePolygonPreviewPolyline`?

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: That makes sense, I added it.

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: I agree with the naming, have changed that. But when it comes to the structure of the file, there are actually the Rectangle Drawing helpers between the Freehand Drawing and it's helpers. I was thinking: first all handlers, then all helpers. (But I had switched the order of freehand and polygon in the helpers part). 

So for me it made most sense to have
- Rectangle Handlers
- Polygon Handlers
- Freehand Handlers
- Rectangle Drawing helpers
- Polygon Drawing helpers
- Freehand Drawing helpers

An alternative is of course to have each helper section below the handler section it belongs to. Would you prefer that?

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: I don't understand which `!` and which variable declaration you are referring to.

packages/@tissuumaps-core/src/controllers/SVGController.ts
cavenel: All good!

packages/@tissuumaps-core/src/controllers/SVGController.ts
cavenel: So, line 372, we have:
```
    const firstPoint = this._freehandDrawingState.points[0]!;
```
with a final `!`. To be consistent, we could do the same here and do:
```
    const firstVertex = this._polygonDrawingState.vertices[0]!;
```
and then remove the final `!` line 272. (Same lines 291 and 195)

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -509,6 +518,53 @@ export class SVGController {
...
+    if (!this._polygonDrawingState) return;
+
+    this._polygonDrawingState.currentPolyline.remove();
+    for (const marker of this._polygonDrawingState.vertexMarkers) {
+      marker.remove();
+    }
+    this._polygonDrawingState = null;
+
+    document.removeEventListener("pointermove", this._handlePolygonPointerMove);
+    document.removeEventListener("keydown", this._handlePolygonKeyDown);
```
cavenel, line 571: Need to add the removeEventListener for `pointercancel` also now

packages/@tissuumaps-core/src/controllers/SVGController.ts
```diff
@@ -320,6 +324,11 @@ export class SVGController {
     }
   };
 
+  private _handlePolygonPointerCancel = (): void => {
```
cavenel: I see now that freehand uses "pointerId" to make sure we cancel on the same pointer id only.
For consistency, we can add `pointerId` in `PolygonDrawingState` (see `_handlePolygonPointerDown`) and check for it in `_handlePolygonPointerCancel`.

cavenel (dismissed):

packages/@tissuumaps-core/src/controllers/SVGController.ts
HenrikeW: I see. Changed it now.

PR #142, opened 2026-05-27
fix: rotate images and layers around origin to match points and shapes

cavenel:

Images rotated around the bottom-right corner instead of the top-left (origin), inconsistent with how points and shapes rotate. Additionally, `flip` was a standalone property on layers and data objects, separate from `transform`, and was applied as its own matrix step rather than being part of the transform itself.

# Type of change

- [x] Refactor (non-breaking change which improves code structure)

# List of changes made

### Rotation origin fix
- Reworked OSD tiled-image geometry in `OpenSeadragonController._updateTiledImageGeometry`: instead of composing/decomposing matrices inline, it now calls the new `TransformUtils.toTiledImageGeometry(transform, layerTransform, contentSize)`, which composes the full data→layer→world matrix (origin-based, exactly like points and shapes) and decomposes it **around the image center**.
- The decomposition is what bridges the two conventions: WebGL/points/shapes rotate around the origin, while OpenSeadragon's `setRotation`/`setFlip` always pivot around the image center. By expressing the composed origin-based transform's translation relative to the image center, the OSD `setPosition` we compute pre-compensates for OSD's internal center pivot, so the net result is a rotation around the origin — consistent with points and shapes.
- Added an optional `center` to `TransformUtils.fromSimilarityMatrix` that performs this translation adjustment, and removed the old `center` rotation-pivot option from `toSimilarityMatrix`.

### Flip moved into Transform
- Added `flip: boolean` to `SimilarityTransform` and renamed the type to `Transform`.
- Flip is now encoded directly in the similarity matrices: `toSimilarityMatrix` applies a `[-1, 1]` scale, and `fromSimilarityMatrix` recovers `flip` from the sign of the 2D determinant — replacing the separate flip matrix step in the controllers.
- Removed the standalone `flip` property from `RawRenderedDataObject` and `RawLayer` (and their defaults).
- Updated `identityTransform` to include `flip: false`.
- Updated `WebGLControllerBase` (`createDataToWorldMatrix` / `createWorldToDataMatrix`) and `OpenSeadragonController` to read flip from `transform.flip`. OSD's `flipped` is now `obj.transform.flip !== layer.transform.flip`.
- Updated `TransformSettingsWidget` to read/write flip directly from the transform, removing the separate `flip`/`onFlipChange` props.
- Removed the now-unnecessary `flip`/`onFlipChange` props from `ImageSettingsWidget`, `LabelsSettingsWidget`, `PointsSettingsWidget`, and `ShapesSettingsWidget`.

# Use of AI

Claude Code (Opus 4.6) was used to investigate the root cause across the codebase.

# Testing

- Load a project with an image layer and a point/shape layer
- Apply rotation at the **element** level (data→layer transform) to both the image and points/shapes — they should both rotate around the top-left corner (origin)
- Apply rotation at the **layer** level (layer→world transform) — images, points, and shapes should all rotate around the origin consistently
- Verify that scale and translation still work correctly in combination with rotation
- Toggle flip on images, points, shapes, and labels — verify it works correctly, and that a layer-level flip mirrors all child objects together as a group
- Load an existing project whose transforms lack `flip` — verify nothing appears flipped by default

# Further comments

**Why the old rotation code was wrong:** the data→layer matrix was built with `center: {x: contentSize.x/2, y: contentSize.y/2}`, correctly producing a rotate-around-center matrix. But the decomposed result is fed to OpenSeadragon's `setPosition`/`setRotation`, and OSD's `setRotation` *also* rotates around the image center internally. The two center offsets compounded rather than cancelled, shifting the effective pivot to the bottom-right corner.

**How it's fixed now:** the transform is composed around the origin (matching points and shapes) and only *decomposed* around the image center via `fromSimilarityMatrix(m, { center })`. That center-aware decomposition converts the origin-based transform into the position/rotation/flip parameters OSD expects, so OSD's internal center pivot is accounted for and the image ends up rotating around the origin (0,0).


packages/@tissuumaps-core/src/controllers/OpenSeadragonController.ts
```diff
@@ -489,15 +489,28 @@ export class OpenSeadragonController {
...
       tiledImageState.ref.layer.transform,
+      {
+        center: {
+          x:
+            (-transform.scale * tiledImageState.tiledImage.getContentSize().x) /
+            2,
+          y:
+            (-transform.scale * tiledImageState.tiledImage.getContentSize().y) /
+            2,
+        },
```
jwindhager: Bit nitpicky, but could we perhaps only compute the `center` value once and use it in both calls?

packages/@tissuumaps-core/src/controllers/OpenSeadragonController.ts
```diff
@@ -489,15 +489,28 @@ export class OpenSeadragonController {
...
     const layerToWorldMatrix = TransformUtils.toSimilarityMatrix(
       tiledImageState.ref.layer.transform,
+      {
+        center: {
+          x:
+            (-transform.scale * tiledImageState.tiledImage.getContentSize().x) /
+            2,
+          y:
+            (-transform.scale * tiledImageState.tiledImage.getContentSize().y) /
+            2,
```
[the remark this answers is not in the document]
cavenel: I believe Copilot has an error in the derivation. After testing, the code appears to be correct.

packages/@tissuumaps-core/src/controllers/OpenSeadragonController.ts
cavenel: And just multiply the pre-computed center by `-transform.scale` in the second one?

jwindhager: Sorry, bit late. Yep, that's what I meant. Beware: after your most recent change, `dataCenter` isn't actually what the name suggests. I'd keep `dataCenter` the actual center and move the negative sign into the function calls.

packages/@tissuumaps-core/src/controllers/OpenSeadragonController.ts
```diff
@@ -484,38 +483,26 @@ export class OpenSeadragonController {
...
     const bounds = tiledImageState.tiledImage.getBoundsNoRotate();
     if (tiledImageState.tiledImage.getFlip() !== flip) {
       tiledImageState.tiledImage.setFlip(flip);
     }
-    const width =
-      tiledImageState.tiledImage.getContentSize().x *
-      dataToWorldTransform.scale;
     if (bounds.width !== width) {
-      tiledImageState.tiledImage.setWidth(width, true); // implicitly updates height to maintain aspect ratio
+      tiledImageState.tiledImage.setWidth(width, true);
```
jwindhager: Why was this comment removed?

packages/@tissuumaps-core/src/model/types.ts
```diff
@@ -58,8 +58,11 @@ export type CoordinateSpace =
   /** World (i.e. global) space */
   | "world";
 
-/** Similarity transform (uniform scale, rotation, and translation) */
-export type SimilarityTransform = {
+/** Similarity transform (uniform scale, rotation, translation, and flip) */
+export type Transform = {
```
jwindhager: As discussed offline, this should not be renamed, as a similarity transform includes flipping. It'd probably be good for future reference to explicitly clarify in the TSDoc comment that this `SimilarityTransform` type explicitly supports direct and indirect (i.e., flipped) similarities.

packages/@tissuumaps-core/src/controllers/WebGLControllerBase.ts
```diff
@@ -20,29 +20,25 @@ export class WebGLControllerBase {
...
+   * flip mirrors the entire coordinate system — all objects move to
+   * mirrored positions as a group.
    *
-   * @param obj - Object providing the flip flag and data → layer transform
+   * @param obj - Object providing the data → layer transform
    * @param layer - Layer providing the layer → world transform
    */
   protected static createDataToWorldMatrix(
-    obj: { flip: boolean; transform: SimilarityTransform },
+    obj: { transform: Transform },
```
jwindhager: I think we could now change this to just passing the `Transform` object itself

packages/@tissuumaps-core/src/controllers/WebGLControllerBase.ts
```diff
@@ -80,31 +76,23 @@ export class WebGLControllerBase {
...
-   * layer → data, then optional horizontal un-flip.
+   * Inverse of {@link createDataToWorldMatrix}
    *
-   * @param obj - Object providing the data → layer transform (inverted) and flip flag
+   * @param obj - Object providing the data → layer transform (inverted)
    * @param layer - Layer providing the layer → world transform (inverted)
    */
   protected static createWorldToDataMatrix(
-    obj: { flip: boolean; transform: SimilarityTransform },
+    obj: { transform: Transform },
```
jwindhager: Same here

packages/@tissuumaps-core/src/controllers/WebGLControllerBase.ts
```diff
@@ -80,31 +76,23 @@ export class WebGLControllerBase {
...
     layer: Layer,
   ): mat3 {
     const worldToDataMatrix = mat3.create();
```
jwindhager, line 88: This line should probably be moved to line 95 (i.e., before the call to `multiply`) for readability & consistency with the change above

packages/@tissuumaps-core/src/utils/TransformUtils.ts
```diff
@@ -1,71 +1,130 @@
...
    */
-  static fromSimilarityMatrix(m: mat3): SimilarityTransform {
+  static fromSimilarityMatrix(
+    m: mat3,
+    options?: { center?: { x: number; y: number } },
+  ): Transform {
     // gl-matrix, like OpenGL, uses column-major order.
+    // Detect reflection via the sign of the 2D determinant.
+    const det = m[0] * m[4] - m[3] * m[1];
+    const flipped = det < 0;
```
jwindhager: Can we rename to `flip` for consistency?

packages/@tissuumaps-core/src/utils/TransformUtils.ts
```diff
@@ -1,71 +1,130 @@
...
    *
    * @param tf - The transform components (all optional)
-   * @param options - Optional rotation center in pre-scaled coordinates
    */
-  static toSimilarityMatrix(
-    tf: Partial<SimilarityTransform>,
-    options?: { center?: { x: number; y: number } },
-  ): mat3 {
-    const { center } = options ?? {};
+  static toSimilarityMatrix(tf: Partial<Transform>): mat3 {
```
jwindhager: Why was the `center` functionality removed from this function?

packages/@tissuumaps-core/src/utils/TransformUtils.ts
```diff
@@ -1,71 +1,130 @@
...
+      center: imageCenter,
+    });
+    return {
+      flip: composed.flip,
+      width: contentSize.x * composed.scale,
+      rotation: composed.rotation,
+      x: composed.translation.x,
+      y: composed.translation.y,
+    };
+  }
```
jwindhager: This is OpenSeadragon-specific and should therefore be part of the `OpenSeadragonController` (if we need a dedicated function for this at all?)

packages/@tissuumaps-core/src/controllers/WebGLControllerBase.ts
```diff
@@ -20,29 +20,25 @@ export class WebGLControllerBase {
...
+   * mirrored positions as a group.
    *
-   * @param obj - Object providing the flip flag and data → layer transform
+   * @param obj - Object providing the data → layer transform
    * @param layer - Layer providing the layer → world transform
    */
   protected static createDataToWorldMatrix(
-    obj: { flip: boolean; transform: SimilarityTransform },
+    obj: { transform: Transform },
     layer: Layer,
```
jwindhager, line 34: Same here

packages/@tissuumaps-core/src/controllers/WebGLControllerBase.ts
```diff
@@ -80,31 +76,23 @@ export class WebGLControllerBase {
...
+   * Inverse of {@link createDataToWorldMatrix}
    *
-   * @param obj - Object providing the data → layer transform (inverted) and flip flag
+   * @param obj - Object providing the data → layer transform (inverted)
    * @param layer - Layer providing the layer → world transform (inverted)
    */
   protected static createWorldToDataMatrix(
-    obj: { flip: boolean; transform: SimilarityTransform },
+    obj: { transform: Transform },
     layer: Layer,
```
jwindhager, line 86: Same here - perhaps we can even get rid of the createDataToWorldMatrix and createWorldToDataMatrix altogether, since all they do is matrix chaining?

jwindhager (changes requested):

packages/@tissuumaps-core/src/utils/TransformUtils.ts
cavenel: Do we need it for anything?

packages/@tissuumaps-core/src/controllers/WebGLControllerBase.ts
cavenel: I think we can keep them, they call toSimilarityMatrix on two transforms and composes them, not so straight forward.

packages/@tissuumaps-core/src/utils/TransformUtils.ts
jwindhager: Not actively. But it was added to `fromSimilarityMatrix`, so I thought for consistency...

packages/@tissuumaps-core/src/controllers/WebGLControllerBase.ts
jwindhager: Ok

apps/tissuumaps/src/components/widgets/TransformSettingsWidget/index.tsx
```diff
@@ -7,29 +7,28 @@ import { Switch } from "@/components/ui/switch";
...
-      )}
+      <Field>
+        <FieldLabel>Flip</FieldLabel>
+        <div className="flex flex-row items-center gap-x-2">
+          <Switch
+            checked={transform.flip}
+            onCheckedChange={(checked) =>
+              onTransformChange({ ...transform, flip: checked })
+            }
+          />
```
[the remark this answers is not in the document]
cavenel: `flip` is a required field of SimilarityTransform, so a transform object without it is simply an invalid project. No backward compatibility at this point so this can be ignored.

jwindhager (approved):
Thanks a lot! Almost there - I just simplified the math a bit and tried to make the intended logic a bit more clear for next when I need to understand transforms 😅 see https://github.com/TissUUmaps/TissUUmaps4/pull/173. After merging #173 we're good to go!

PR #146, opened 2026-06-10
fix: Always add dummy background image covering point data bounds

cavenel:
Previously the transparent dummy image was only created when no real images were loaded. This caused the viewport to be locked to a tiny real image when point data extended far beyond it. The dummy is now always added so OpenSeadragon's world bounds cover all points.

Jira issue: TMAP-341

# Type of change

# List of changes made

- Always create the transparent dummy image when point data bounds exist, regardless of whether real images are loaded. Previously it was only created when no real images were present.
- Update dummy position and size on subsequent synchronize calls when bounds change, instead of only creating it once.
- Only fit viewport to dummy bounds when there are no real images, preserving existing viewport behavior when a real image is loaded.
- Remove the dummy only when there is no point data at all (`dummyBounds` is null).

# Screenshot of the fix

N/A

# Use of AI

This PR was created with the assistance of Claude Code (Claude Opus 4.6).

# Testing

- Load a project with a small real image and point data that extends far beyond the image bounds
- Verify the viewport can pan and zoom to reach all point data
- Load a project with no real image and only point data — verify it still fits to the points as before
- Load a project with a normal-sized image and point data within the image — verify no regression

# Further comments

N/A

### DIFF
--- a/packages/@tissuumaps-core/src/controllers/OpenSeadragonController.ts
+++ b/packages/@tissuumaps-core/src/controllers/OpenSeadragonController.ts
@@ -526,7 +526,6 @@ export class OpenSeadragonController {
 
   private _updateDummy(dummyBounds: Rect | null): void {
     if (dummyBounds !== null && this._dummyTiledImage === undefined) {
-      const fitBounds = this._tiledImageStates.length === 0;
       const index = this.viewer.world.getItemCount();
       this.viewer.addTiledImage({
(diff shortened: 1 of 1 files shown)


jwindhager (approved):
lgtm, good work!

Could you please add a corresponding Jira issue for this and update the PR description? Just to keep track of the work. Thanks :)

PR #148, opened 2026-06-11
perf(storage): parse data off the main thread (CSV/Parquet/GeoJSON workers)

jwindhager:

Jira: **TMAP-150**

Follow-up to #145 (which moved CPU-side marshalling off the critical path); this PR tackles the remaining storage-adapter **parsing**.

# Type of change

- [ ] Other

# List of changes made

Data parsing in the storage adapters ran synchronously on the main thread, freezing the UI while large datasets loaded. This PR moves it off-thread, preferring the parser libraries' own worker support and adding custom workers only where needed.

- **CSV** (`CSVTableDataProvider`): enable papaparse's built-in worker (`worker: true`) so tokenization runs off-thread; drive parsing from the batched `chunk` callback instead of per-row `step`. papaparse falls back to a main-thread parse where `Worker` is unavailable.
- **Parquet** (`parquet.worker.ts`, `ParquetWorkerClient`, `ParquetTableData`): a custom `?worker&inline` worker owns the `ArrayBuffer` and runs all hyparquet decoding (metadata, id/name, lazy per-column reads); typed column buffers transfer zero-copy. hyparquet + hyparquet-compressors leave the main bundle entirely (bundled only into the self-contained worker blob).
- **GeoJSON** (`geojson.worker.ts`, `geojsonParse`): a custom `?worker&inline` worker runs `JSON.parse` + the coordinate transform and returns a flat, transfer-friendly `ShapesGeometry` (struct-of-arrays with CSR offsets; coords as `Float32Array`).
- **Geometry representation:** add `ShapesGeometry` to `@tissuumaps/core`; change `ShapesData.loadMultiPolygons` → `loadGeometry`; rewrite `WebGLShapesController._getDataBounds` / `_createScanlines` to walk the flat offsets (rasterization output unchanged). `MultiPolygon` is retained for the interactive shape-drawing path.
- AbortSignal cancellation is preserved across all worker boundaries.

# Screenshot of the fix

N/A.

# Use of AI

Yes. Planned and implemented with Claude Code (Claude Opus 4.8) in an agent session; the diff and tests were reviewed before committing.

# Testing

A rebuild is needed due to new dependencies: **No** (no new dependencies).

Automated:
- `pnpm --filter @tissuumaps/core test` and `pnpm --filter @tissuumaps/storage test` (incl. the flat-geometry conversion and the Parquet RPC client).
- `pnpm build` confirms the production single-file `index.html` is intact with no separate worker chunks, and that hyparquet/OpenSeadragon are not pulled into the worker bundles.

Manual (recommended before merge): `pnpm dev`, then load a large CSV, Parquet, and GeoJSON and confirm the UI stays interactive during loading and the rendered points/shapes match `development`.

# Further comments

Draft: opening for early review while the manual `pnpm dev` verification above is completed. Targets `development`. The storage package's unit tests run in node and cannot import the `@tissuumaps/core` barrel (it eagerly loads OpenSeadragon, which needs `document`), so worker runtime paths and shape rasterization equivalence are covered by build + typecheck rather than unit tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/tissuumaps/src/store/adapters/LoadedShapesDataAdapter.ts
+++ b/apps/tissuumaps/src/store/adapters/LoadedShapesDataAdapter.ts
@@ -1,9 +1,9 @@
 import { useCallback } from "react";
 
 import {
-  type MultiPolygon,
   type ProgressCallback,
   type ShapesData,
+  type ShapesGeometry,
 } from "@tissuumaps/core";
 import { type ViewerAdapter } from "@tissuumaps/viewer";
 
@@ -28,14 +28,14 @@ export class LoadedShapesDataAdapter implements ShapesData {
     return this._getData().getNames();
   }
 
-  async loadMultiPolygons(options?: {
+  async loadGeometry(options?: {
     signal?: AbortSignal;
     onProgress?: ProgressCallback;
-  }): Promise<MultiPolygon[]> {
+  }): Promise<ShapesGeometry> {
     const { signal, onProgress } = options ?? {};
     signal?.throwIfAborted();
     const state = useTissUUmaps.getState();
-    return await state.loadShapesMultiPolygons(this._shapesId, {
+    return await state.loadShapesGeometry(this._shapesId, {
       signal,
       onProgress,
     });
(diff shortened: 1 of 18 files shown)


jwindhager:
Closing in favor of https://github.com/TissUUmaps/TissUUmaps4/pull/151 and following

PR #154, opened 2026-06-22
Fix/data source widget validation

cavenel:

Follow-up to the data source type switcher introduced in #138.
Jira: TMAP-350

# Type of change

- [ ] Other

# List of changes made

- Validate the data source draft directly against the active provider's JSON schema in `DataSourceWidget`, instead of relying on the `hasErrors` state fed by JsonForms' `onChange`. The old state was reset to `false` on type switch / reset / when entering edit mode, so the Save button could be enabled while required fields of the newly selected provider were still unsatisfied.
- Compile the schema with `createAjv()` (the same Ajv configuration JsonForms uses internally), so validation results match the form and shared fields (e.g. `url`) carry over across a type switch.
- Only run validation while editing (`isEditing && !validate(...)`), since the result is only used to gate the Save button.

# Screenshot of the fix

N/A — no visual change; only the Save button's enabled/disabled state is affected.

# Use of AI

AI (Claude Code) was used to review the change and to apply a small refactor. The fix was authored and verified by me.

# Testing

Instructions on how to test:

1. Open a data object's `DataSourceWidget` and click Edit.
2. Switch the source type to a provider with required fields that the current draft doesn't satisfy (e.g. one requiring `url`).
3. Confirm the Save button stays disabled until all required fields for the selected type are filled in.
4. Confirm that shared fields (e.g. `url`) carry over across the type switch and that Save re-enables once the draft is valid.

# Further comments

Minimal change scoped to `DataSourceWidget`; no API or schema changes.


jwindhager (approved):
lgtm

PR #162, opened 2026-07-22
refactor(render)!: extract rendering into a dedicated @tissuumaps/render package

jwindhager:

None.

# Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) — packages are now ESM-only (see below)
- [x] Other — large internal refactor / code reorganization

# List of changes made

The bulk of this PR is a refactor, but it also carries deliberate functional changes — see
the dedicated section below. Structural changes:

- **Extracted rendering into a new `@tissuumaps/render` package.** All WebGL and SVG
  rendering, previously living in `@tissuumaps/core` (and partly in the viewer), now has a
  dedicated home with a clear public surface (see `packages/@tissuumaps-render/src/index.ts`).
  This keeps `core` free of rendering/OpenSeadragon concerns.
- **Renamed the "Controller" abstraction to "Renderer" and restructured it.** The old
  `core/src/controllers/*` (`OpenSeadragonController`, `WebGLController`,
  `WebGLPointsController`, `WebGLShapesController`, `WebGLControllerBase`) are gone, replaced
  by renderers under `@tissuumaps-render`: `WebGLPointsRenderer`, `WebGLShapesRenderer`,
  `WebGLRendererBase`, `OpenSeadragonImageRenderer`, `OpenSeadragonLabelsRenderer`,
  `OpenSeadragonRendererBase`, plus new `WebGLContext`/`OpenSeadragonContext` and
  `WebGLOptions`/`OpenSeadragonOptions` option types. `WebGLControllerBase` became
  `webgl/WebGLUtils`.
- **Moved the per-attribute data resolvers** out of `core/src/utils/data/*DataUtils.ts` into
  `@tissuumaps-render/src/webgl/resolvers/*Resolver.ts` (`Color`, `Marker`, `Opacity`, `Size`,
  `Visibility`, and `DataUtilsBase` → `ResolverBase`), matching their renderer-facing role.
- **Added a `WebGLShapesRasterizer`** (with tests) to encapsulate shape rasterization.
- **Reorganized `@tissuumaps/core` types.** `src/types.ts` and `model/types.ts` were split into
  focused modules: `types/arrays.ts`, `types/callbacks.ts`, `types/geometry.ts`,
  `types/legacy.ts`, and `model/primitives.ts`. Added `GeometryUtils` (with tests).
- **Build/tooling housekeeping:** introduced per-package `tsconfig.src.json`, renamed
  `tsconfig.dts.json` → `tsconfig.ts59.json`, updated `vite.config.ts` files and root tsconfig
  project references, and regenerated `pnpm-lock.yaml` for the new package.

# Functional changes (not pure restructuring)

- **Dropped non-ESM (CommonJS/UMD) build output — packages are now ESM-only.** Every package
  (`core`, `storage`, `viewer`, `plugins`, and the new `render`) now builds with Vite
  `formats: ["es"]` only; the `"umd"`/`"cjs"` lib formats and UMD global names were removed.
  Correspondingly, `package.json` drops the top-level `main`/`module`/`types` fields and the
  `require` export condition (replaced by `"default": "./dist/*.js"`). **This breaks any
  downstream consumer relying on `require()` / CommonJS / a UMD global** — consumers must now
  use ESM `import`.
- **Restructured the OpenSeadragon rendering strategy into two anchored renderers.** The single
  `OpenSeadragonController` was split into `OpenSeadragonImageRenderer` and
  `OpenSeadragonLabelsRenderer` (both extending `OpenSeadragonRendererBase`). Each renderer owns
  a transparent, zero-opacity *anchor* tiled image placed at a fixed z-index (`anchorIndex`);
  the renderer's object tiled images then occupy the contiguous "tiled-image band" of indices
  immediately after their anchor. `cleanRenderedObjects` enforces this band — each surviving
  object must sit at the expected consecutive index (`anchorIndex + 1 + n`), otherwise it is
  removed and recreated — and the anchor is continuously resized to the bounding box of its band
  (`updateBounds`). This gives images and labels independent, stable slices of the OSD world
  z-order that cannot interfere with one another. Relatedly, the old polled "dummy bounds"
  fallback was replaced by a push model: the WebGL renderers now return their rendered bounds
  from `synchronize()`, which `Viewer` feeds into the OSD renderers via `setExtraBounds` so the
  world/home bounds extend to cover off-image points/shapes.
- **Serialized OpenSeadragon world-item mutations (fixes HMR).** `OpenSeadragonContext` now
  funnels every mutation of the viewer world — `addTiledImage`, `removeTiledImage`,
  `getTiledImageIndex`, `updateBounds` — through a single FIFO promise chain
  (`_enqueueItemsMutation` / `_itemsMutationPromise`), and `destroy()` awaits any pending
  mutation before tearing the viewer down. Because OSD's `addTiledImage` is asynchronous and
  FIFO, previously the rapid effect teardown/re-init during hot module replacement could
  interleave add/remove operations and leave the world in a corrupt state (orphaned/duplicate
  tiled images, wrong ordering); serializing them makes HMR reliable.
- **WebGL context-loss handling and forced GPU-resource release.** `useWebGL` now listens for
  `webglcontextlost`/`webglcontextrestored` and stops/restarts the renderers, and
  `WebGLContext.destroy()` force-loses the context via the `WEBGL_lose_context` extension to
  immediately free GPU resources — so repeated remounts under React StrictMode/HMR don't exhaust
  the browser's live-WebGL-context cap. This is the WebGL counterpart to the OSD
  mutation-serialization fix above.
- **Reworked the viewer overlay hooks (`useSVG`, `useWebGL`, `useOpenSeadragon`) and `Viewer`.**
  Initialization moved from a `parent` + `initialViewport` effect to a callback-ref pattern
  (e.g. `initSVG`), and the hooks now reactively sync live `viewport`, `containerSize`, and
  interaction-mode changes to the underlying controller/renderer instead of re-initializing.
  The `OpenSeadragonController` context/provider was also renamed to `OpenSeadragonContext`.
  This changes the overlay lifecycle and update timing.
- **Adopted a "callee owns the abort signal" convention.** Redundant `signal.throwIfAborted()`
  calls in promise success handlers and after worker round-trips were removed (e.g. in
  `ParquetTableData.loadValues`/`loadValueRange`); remaining checks were kept and commented
  where they guard subsequent O(n) work. This changes where/when aborts are detected at runtime.
- **Normalized async signatures** (e.g. `GeoJSONShapesData.loadGeometry` and the parquet
  worker's `slice` are now `async`; sync-body `return await Promise.resolve(x)` simplified to
  `return Promise.resolve(x)`) — behavior-neutral, but part of the same async harmonization.

# Screenshot of the fix

N/A.

# Use of AI

AI (Claude Code) was used both in developing the code changes on this branch and in drafting
this PR description (the latter by inspecting the branch diff).

# Testing

- A rebuild is needed due to new dependencies (new `@tissuumaps/render` workspace package; run `pnpm install`).
- Instructions on how to test: build all packages and run the test suites (`pnpm -r build`, `pnpm -r test`); tests for the moved resolvers, WebGL utils, the shapes rasterizer, and geometry utils travel with the code. Then run the app and confirm images, points, and shapes still render, that the SVG/WebGL overlays track viewport pan/zoom and container resizes (given the hook rework), and that hot module replacement / StrictMode remounts no longer corrupt the OSD world or exhaust WebGL contexts.

# Further comments

Most of the diff is file moves/renames (rendering out of `core`) plus import and build-config
updates, but reviewers should pay attention to the "Functional changes" section above — in
particular the **ESM-only packaging** change (breaks CommonJS/UMD consumers), the viewer
overlay hook rework, and the HMR/context-loss handling.


jwindhager:
Sorry for the enormous PR. I promise future PRs will be smaller. Also, note that the current recreate-if-index-changed strategy of the OSD renderers is a workaround for the `World.setItemIndex` not playing well with the minimap (?); need to investigate this (in a future PR 😇). Finally, `fitBounds` is currently always called when world bounds change (e.g., when new data loads) - this needs fixing (users may change the viewport while data is still loading, or may have changed it already before loading new data).

PR #170, opened 2026-09-01
feat(image)!: multi-channel image rendering with per-channel color and opacity

jwindhager:

[TMAP-162](https://scilifelab.atlassian.net/browse/TMAP-162) — Add multi-channel image support.

Builds on the incremental renderer synchronization from #167. **No UI is wired up yet** (see *Further comments*).

# Type of change

- [ ] Other

# List of changes made

Renders multi-channel images as one OpenSeadragon TiledImage per channel, additively blended onto an opaque backdrop, with a per-channel color, visibility and opacity taken from the model.

**Model (`@tissuumaps/core`)**

- New `Channel` type in `model/image.ts` (`name?`, `visibility?`, `opacity?`, `color?`), and `RawImage.channels?: Channel[]`, indexed by channel. Channels beyond the end of the array, and images whose data is not multi-channel, fall back to the defaults (visible, opacity 1, no tint). `RawImage` is no longer an empty interface, so the `no-empty-object-type` suppression is gone.
- `ImageData` (`storage/image.ts`) becomes explicitly two-shaped: it either provides one tile source *per channel*, or a single unaddressed one. `getSizeC()` returns `undefined` in the latter case; `getTileSource(c?)` takes a channel index that is required for multi-channel data, must be omitted otherwise, and throws when it is out of bounds. Channel metadata is exposed through four *optional* per-channel accessors — `getChannelName?(c)`, `getChannelVisibility?(c)`, `getChannelOpacity?(c)` and `getChannelColor?(c)` — replacing the whole-image `getChannelNames()`, so a data provider can report what a format already knows (OME-Zarr `omero` rendering settings, for instance) without every implementation having to grow four methods.
- `OpenSeadragonViewerOptions` now also omits `compositeOperation` — viewer options are applied to *every* tiled image in the world, which would overwrite the per-tiled-image composite operations the renderers now set.
- `ColorUtils.colorsEqual()` (exact component comparison, no tolerance) plus 4 tests, used to skip redundant tint invalidations.

**Tinting (`OpenSeadragonContext`)**

- New `setTiledImageTint(tiledImage, tint | undefined)`. OpenSeadragon has no per-image color, so the tint is applied to the *tiles*, in a `tile-invalidated` handler: the tile is composited over opaque black (`destination-over`, which turns transparency into intensity), then `multiply`-ed with the tint.
- Tints are stored in a `WeakMap` keyed by **tile source**, not tiled image, because that is the granularity at which OSD keys its tile caches — and it also covers the navigator, whose tiled images mirror the world but share its tile sources. Consequence, documented on the method: tiled images sharing a tile source share a tint, last write wins.
- The tint is only written and the tiles only invalidated when the color actually changed; `requestInvalidate(restoreTiles: true, viewportOnly: false)` re-runs the tinting from the original tile data on every loaded tile.
- `_openTileSource()` is now public `openTileSource()` and additionally awaits a *promised* tile source. This lets a caller derive one tile source from another that is still opening without losing its place in the addition queue (and therefore its world index) — which is exactly what the backdrop needs.

**Per-channel tiled images (`OpenSeadragonRendererBase`)**

- `RenderedObject` holds `tiledImages: TiledImage[]` and an optional `backdrop` instead of a single `tiledImage`; `tiledImagePromise` → `tiledImagesPromise`. `tileSourceCount` is recorded up front, at creation time, because the world layout has to be computable before the images exist.
- An object now occupies `tileSourceCount` consecutive world indices after the anchor, optionally preceded by its backdrop. `createRenderedObject(offset, …)`'s `offset` therefore counts **TiledImages, not objects**; callers advance it by `tileSourceCount` (+1 for a backdrop). `cleanRenderedObjects()` checks that the backdrop *and* every tiled image sit at their expected consecutive index — a partially misplaced object is not reusable.
- All additions for an object are requested before `createRenderedObject()` returns and in world order, then joined with `Promise.allSettled`. If any one of them fails, the whole object is torn down again: a partially added object would shift the world indices of everything behind it. Deletion removes tiled images and backdrop together.
- New overridable `usesAdditiveBlending(data)` (default `false`), `getTiledImageColor(ref, c)` (default `undefined`) and `getTiledImageOpacity(ref, c?)` (default: the old layer × object opacity, `0` when either is invisible). `getOpacity`/`getTransform` were renamed to `getTiledImageOpacity`/`getTiledImageTransform`; the former stops being `static` since subclasses now override it. The per-image write logic moved into a private `_updateTiledImage()`, still writing only properties whose value actually changed.
- The backdrop is an opaque-black single-pixel tile source sized to the *first opened* channel's dimensions, added with `opacity: 0` (made visible only once transformed) and, deliberately, **no explicit `compositeOperation`** — setting it, even to `"source-over"`, would exclude the backdrop from the WebGL drawer's batched path. The channels themselves get `compositeOperation: "lighter"`, so they add up among themselves while the backdrop hides what is below, compositing the object as a whole over the rest of the world.

**Image renderer (`OpenSeadragonImageRenderer`)**

- `getTileSource` → `getTileSources`, returning one tile source per channel for multi-channel data and a single one otherwise (`OpenSeadragonLabelsRenderer` follows the renamed signature; it still throws).
- Overrides `usesAdditiveBlending` (true iff `getSizeC() !== undefined`), `getTiledImageColor` (the channel's color; `undefined` leaves the tiles in their own colors) and `getTiledImageOpacity` (multiplies the base opacity by the channel's visibility and opacity; the backdrop, called without `c`, keeps the plain layer × object opacity). Each of the three channel properties is resolved model first, data second, default last: an explicit value on `Image.channels[c]` wins, otherwise the data's optional accessor is consulted, otherwise the default (visible, opacity 1, no tint) applies.
- `synchronize()` now walks `newRefs` with a running TiledImage offset instead of using the loop index, and images whose tiled images cannot be created — e.g. data that yields no tile sources — are logged and skipped, like images whose data failed to load.

**Storage providers**

- `OMEZarrImageDataSource.c` (a single channel to show) is replaced by `sizeC` (the number of channels to open). `OMEZarrImageDataProvider` builds one `OMEZarrTileSource` per channel when `sizeC` is set and a single one otherwise, and the URL/object-URL resolution was flattened into one `if`/`else if` chain so both paths share the tile-source construction. The "Channel" control is gone from the uischema.
- `OpenSeadragonImageData` reports "not multi-channel" (`getSizeC()` → `undefined`) and throws if a channel index is passed. `ImageDataWrapper` forwards `getSizeC()`, the channel-indexed `getTileSource()` and the four optional per-channel accessors.

**Misc**

- `OpenSeadragonUtils` (new) holds `transparentPixelUrl`, the new `blackPixelUrl`, and `createPixelTileSource(size, pixelUrl)`, extracted from the anchor tile source in `OpenSeadragonContext` and now shared with the backdrop.
- `tiledImage.update(true)` on a 0 → visible opacity transition became `update(/* viewportChanged */ false)`; the argument is `viewportChanged`, and the viewport has not changed here.

# Screenshot of the fix

N/A for now — nothing in the UI exposes channels yet, so there is no user-visible change to screenshot (see *Further comments*).

# Use of AI

Yes. As with the rest of `development`, the code was written with the assistance of Claude Code (Opus 5) and manually reviewed; Claude Code also drafted this description and ran the verification below.

# Testing

- Instructions on how to test

`pnpm run build`, `eslint .`, `prettier --check .` and `pnpm test` (439 passed / 18 files) all pass on this branch. Only the 4 new `ColorUtils.colorsEqual` tests are new — the OpenSeadragon renderers have no test coverage, so the rendering itself has to be exercised by hand.

To do that, an OME-Zarr image data source with `sizeC` set has to be put into a project file directly, together with `channels` on the image (see *Further comments*). Worth checking: channels adding up rather than occluding each other; toggling a channel's `visibility`/`opacity`/`color`; single-channel and non-OME-Zarr (DZI/IIIF) images still rendering unchanged with their transparency intact; the navigator showing the same tints; and reordering/removing images, which is where the consecutive world indices are easy to get wrong.

# Further comments

**No UI, and no channel discovery.** `Image.channels` and the `ImageData` channel accessors are model and storage plumbing only — nothing reads `channels` outside the renderer, and nothing writes it, so channels can currently only be configured by hand in a project file. No data provider implements the optional accessors yet either: `OMEZarrImageDataProvider` does not read channel metadata at all, `sizeC` has to be supplied by the caller, and channel names are left `undefined` (`// TODO channel names`). Both are follow-up work.

**Tinting drops tile transparency.** Compositing over black is what makes `"lighter"` blending behave, but it means a tinted tiled image *must* sit on the backdrop; this is why only multi-channel image data is tinted, and why `getTiledImageColor()` returns `undefined` for everything else.

### DIFF
--- a/packages/@tissuumaps-core/src/model/image.ts
+++ b/packages/@tissuumaps-core/src/model/image.ts
@@ -13,12 +13,28 @@ import {
 export const imageDefaults = {} as const satisfies Partial<RawImage>;
 
 /**
+ * A channel of a two-dimensional raster image
+ */
+export type Channel = {
+  /** The name of the channel (overrides data provider) */
+  name?: string;
+
+  /** The visibility of the channel */
+  visibility?: boolean;
+
+  /** The opacity of the channel */
+  opacity?: number;
+};
+
+/**
  * A two-dimensional raster image
  */
-// eslint-disable-next-line @typescript-eslint/no-empty-object-type
 export interface RawImage extends RawSingleLayerDataObject<
   RawImageDataSource<string>
-> {}
+> {
+  /** The channels of the image */
+  channels?: Channel[];
+}
 
 /**
  * A {@link RawImage} with {@link imageDefaults} applied
--- a/packages/@tissuumaps-core/src/model/image.ts
+++ b/packages/@tissuumaps-core/src/model/image.ts
@@ -14,15 +14,28 @@ export const imageDefaults = {} as const satisfies Partial<RawImage>;
 
 /**
  * A channel of a two-dimensional raster image
+ *
+ * Channels are only applied to multi-channel image data, as reported by the
+ * image's data provider.
  */
 export type Channel = {
-  /** The name of the channel (overrides data provider) */
+  /**
+   * Channel name, overriding the name reported by the data provider
+   */
   name?: string;
 
-  /** The visibility of the channel */
+  /**
+   * Channel visibility
+   *
+   * @defaultValue `true`
+   */
   visibility?: boolean;
 
-  /** The opacity of the channel */
+  /**
+   * Channel opacity, in the range [0, 1], multiplied with the image opacity
+   *
+   * @defaultValue `1`
+   */
   opacity?: number;
 };
 
@@ -32,7 +45,12 @@ export type Channel = {
 export interface RawImage extends RawSingleLayerDataObject<
   RawImageDataSource<string>
 > {
-  /** The channels of the image */
+  /**
+   * The channels of the image, indexed by channel
+   *
+   * Channels beyond the end of the array, and channels of images whose data is
+   * not multi-channel, use the default values of {@link Channel}.
+   */
   channels?: Channel[];
 }
 
--- a/packages/@tissuumaps-core/src/model/image.ts
+++ b/packages/@tissuumaps-core/src/model/image.ts
@@ -6,6 +6,7 @@ import {
   createDataSource,
   createSingleLayerDataObject,
 } from "./base";
+import type { Color } from "./primitives";
 
 /**
  * Default values for {@link RawImage}
@@ -37,6 +38,13 @@ export type Channel = {
    * @defaultValue `1`
    */
   opacity?: number;
+
+  /**
+   * Channel color, multiplied with the channel's image data
+   *
+   * Without a color, the channel's image data is rendered in its own colors.
+   */
+  color?: Color;
 };
 
 /**
--- a/packages/@tissuumaps-storage/src/ome-zarr/OMEZarrImageDataProvider.ts
+++ b/packages/@tissuumaps-storage/src/ome-zarr/OMEZarrImageDataProvider.ts
@@ -25,7 +25,7 @@ export class OMEZarrImageDataProvider implements ImageDataProvider<
         type: "string",
       },
       // TODO path
-      c: {
+      sizeC: {
         type: "integer",
       },
       z: {
@@ -47,6 +47,11 @@ export class OMEZarrImageDataProvider implements ImageDataProvider<
         label: "URL",
       },
       // TODO path
+      {
+        type: "Control",
+        scope: "#/properties/sizeC",
+        label: "Number of channels",
+      },
       {
         type: "Control",
         scope: "#/properties/z",
--- a/packages/@tissuumaps-storage/src/ome-zarr/OMEZarrImageDataProvider.ts
+++ b/packages/@tissuumaps-storage/src/ome-zarr/OMEZarrImageDataProvider.ts
@@ -95,17 +95,24 @@ export class OMEZarrImageDataProvider implements ImageDataProvider<
     } else {
       throw new Error("A URL or workspace path is required to load data.");
     }
-    const tileSources = [];
+    let tileSource: OMEZarrTileSource | undefined;
+    let tileSources: OMEZarrTileSource[] | undefined;
+    let channelNames: string[] | undefined;
     const { sizeC, z, t } = normalizedDataSource;
-    if (sizeC !== undefined) {
-      for (let c = 0; c < sizeC; c++) {
-        const tileSource = new OMEZarrTileSource({ url, c, z, t });
-        tileSources.push(tileSource);
-      }
+    if (sizeC === undefined) {
+      tileSource = new OMEZarrTileSource({ url, z, t });
     } else {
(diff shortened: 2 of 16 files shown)


packages/@tissuumaps-core/src/model/image.ts
```diff
@@ -6,19 +6,61 @@ import {
...
  */
 export const imageDefaults = {} as const satisfies Partial<RawImage>;
 
+/**
+ * A channel of a two-dimensional raster image
+ *
+ * Channels are only applied to multi-channel image data, as reported by the
+ * image's data provider.
+ */
+export type Channel = {
```
cavenel, line 22: Should channel also be re-exported from src/index.ts? Like the other model types (Image, RawImage, etc.)

packages/@tissuumaps-storage/src/ome-zarr/OMEZarrImageDataProvider.ts
```diff
@@ -84,35 +84,35 @@ export class OMEZarrImageDataProvider implements ImageDataProvider<
...
     }
-
-    throw new Error("A URL or workspace path is required to load data.");
+    let tileSource: OMEZarrTileSource | undefined;
+    let tileSources: OMEZarrTileSource[] | undefined;
+    const { sizeC, z, t } = normalizedDataSource;
+    if (sizeC === undefined) {
+      tileSource = new OMEZarrTileSource({ url, z, t });
+    } else if (sizeC <= 0) {
+      throw new Error("Number of channels must be a positive integer.");
```
cavenel: When sizeC <= 0 throws, the URL.createObjectURL(file) from line 94 is never revoked. OMEZarrImageData.close() owns revocation but is never constructed. sizeC comes from normalizedDataSource, so the check could move to the top of openData, before any file handling?

cavenel (commented):
Two minor comments and then good to merge I think!

packages/@tissuumaps-core/src/model/image.ts
jwindhager: Good catch! Could you maybe fix it? 😇

packages/@tissuumaps-storage/src/ome-zarr/OMEZarrImageDataProvider.ts
jwindhager: Also a good catch! Sounds good, could you maybe implement it? 😇

PR #171, opened 2026-09-01
feat(storage)!: resolve relative URLs against the project URL

jwindhager:

[TMAP-354](https://scilifelab.atlassian.net/browse/TMAP-354)

# Type of change

- [ ] Other

# List of changes made

- Relative `url`s in data sources are now resolved against the URL the project was loaded from - after any redirects - instead of the document base URL, so that a project directory can be served from any subdirectory or host without editing the project file. Projects opened from a local file have no such base and keep resolving against the document base URL.
- The project URL is kept in the project store (`ProjectStoreState.url`) and written in the same update as the project itself, so that the two can never disagree. It describes where the project came from, not what it contains: `cleanProject()` drops it, and every save path now goes through `saveProject()`, so it can never end up in a saved `.tmap`.
- The data caches take the project URL as part of their context. It is part of the memoized data source resolution key and triggers cache reconciliation, so that entries are re-normalized and reloaded when it changes.
- `DataProvider.normalizeDataSource(dataSource)` becomes `normalize(dataSource, projectUrl)`, and `load()` now takes the already normalized data source rather than normalizing internally. The provider interfaces gained a `TNormalizedDataSource` type parameter (defaulting to `TDataSource`), and the `Default*DataSource` types were renamed to `Normalized*DataSource`.
- Documented the resolution rules in the projects concept page, in the `url` docstring and in the cache section of the architecture docs, and added a regression test for re-normalization on a project URL change.

# Screenshot of the fix

n/a

# Use of AI

Claude Code (Opus 5) was used during development: implementing parts of the change, reviewing the diff over several rounds, and writing docstrings, documentation and the added test.

# Testing

- Instructions on how to test:
  - `pnpm build && pnpm test` (440 tests).
  - Serve a project file with a relative data source `url` from a subdirectory and open it with `?project=<subdir>/project.tmap`: the data is fetched from that subdirectory, not from the application root. Loading the same file through the project panel's file picker still resolves it against the application root.

# Further comments

- Behavioral change for existing projects: a project served next to the application is unaffected, but one served from a subdirectory now resolves its relative URLs within that subdirectory. Absolute URLs are unaffected. This is called out in the projects concept page.
- Known limitation, left as is: a malformed `url` makes `normalize()` throw out of the project store listener, rather than surfacing as an error data ref for the affected object.


jwindhager:
As discussed offline, this should also work for project.json files below an open workspace, but we can probably address this separately in another PR.

packages/@tissuumaps-core/src/model/base.ts
```diff
@@ -235,7 +235,11 @@ export interface RawDataSource<TType extends string = string> extends RawModel {
   type: TType;
 
   /**
-   * Remote URL (absolute or relative to TissUUmaps root)
+   * Remote URL, absolute or relative
+   *
+   * Relative URLs are resolved against the URL the project was loaded from,
+   * and against the TissUUmaps root for projects without one (i.e. projects
```
cavenel, line 241: Maybe we could unify naming convention a bit? The model doc says "TissUUmaps root", the provider doc in storage/base.ts says "document base URL", and apps/docs/docs/concepts/projects.md says "TissUUmaps application root". The code uses document.baseURI, which is the document base URL. Right?

cavenel (approved):
One minor doc comment that can be ignored for now. Rest looks great!

packages/@tissuumaps-core/src/model/base.ts
jwindhager: Yep 🙃

PR #173, opened 2026-09-04
refactor(core): pivot-based similarity matrix decomposition

jwindhager:

Follow-up to `fix-transform`. No associated JIRA issue.

# Type of change

- [ ] Other

# List of changes made

- `TransformUtils.fromSimilarityMatrix`: replace `options.center` with a positional `pivot` and compute the pivoted translation as `m(pivot) - scale * pivot` directly from the matrix columns, dropping the atan2 → cos/sin round trip; use `Math.hypot` for the scale.
- `TransformUtils.toSimilarityMatrix`: add a matching optional `pivot` argument so it is the exact inverse of `fromSimilarityMatrix` for the same pivot; flip/scale are now applied together before rotation.
- Extend `TransformUtils` tests with pivot round-trips and OpenSeadragon-convention cases.
- `render`: adopt the new `pivot` argument in `OpenSeadragonUtils`, simplify `mat3.multiply` usage in `WebGLUtils`, fix TSDoc.

# Screenshot of the fix

N/A

# Use of AI

Claude Code was used in development.

# Testing

- `pnpm test` in `@tissuumaps/core` (TransformUtils tests)

# Further comments

None.

### DIFF
--- a/packages/@tissuumaps-core/src/utils/TransformUtils.ts
+++ b/packages/@tissuumaps-core/src/utils/TransformUtils.ts
@@ -50,26 +50,45 @@ export class TransformUtils {
   /**
    * Builds a 3×3 similarity matrix from a (partial) {@link SimilarityTransform}
    *
-   * Applies, in order: flip, scale, rotation, and translation.
+   * Applies, in order: flip/scale, rotation, and translation.
+   *
+   * If `pivot` is given, flip and rotation are applied about `pivot` (in the
+   * source coordinates of the transform) instead of the origin, while scale
+   * stays about the origin. The translation is then the position of the
+   * scaled content before it is flipped and rotated about its (scaled)
+   * pivot. This is the inverse of {@link fromSimilarityMatrix} called with
+   * the same pivot.
    *
    * @param tf - The transform components (all optional)
+   * @param pivot - Optional point about which flip and rotation are applied
    * @returns The composed matrix
    */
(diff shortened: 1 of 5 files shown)


packages/@tissuumaps-core/src/utils/TransformUtils.ts
```diff
@@ -15,72 +15,81 @@ export class TransformUtils {
...
       mat3.rotate(m, m, (Math.PI * tf.rotation) / 180);
     }
-    if (tf.scale !== undefined) {
-      mat3.scale(m, m, [tf.scale, tf.scale]);
-    }
     if (tf.flip) {
       mat3.scale(m, m, [-1, 1]);
     }
+    if (pivot !== undefined) {
+      const scale = tf.scale ?? 1;
```
cavenel, line 87: It computes scale twice, not a huge effort I guess but could we simplify?

cavenel (approved):
Ok to add the pivot in `toSimilarityMatrix`, it's fine for consistency! And better with readability, so happy to merge!

packages/@tissuumaps-core/src/utils/TransformUtils.ts
jwindhager: Sure! It's just a ??, but happy to move it up. Feel free to do it yourself if you dont want to wait until I'm near my laptop :-)

PR #174, opened 2026-09-07
Fix/viewport sync draw

cavenel:

Reintroduced by #162 (`refactor(render)!: extract rendering into a dedicated @tissuumaps/render package`), which rewrote the `viewport-change` handler to store the viewport in React state. Same class of bug as the overlay lag fixed once before.

# Type of change

- [ ] Other

# List of changes made

The WebGL (points, shapes) and SVG overlays lagged one frame behind the OpenSeadragon canvas on pan/zoom. The `viewport-change` handler in `Viewer` wrote the viewport into React state; `useWebGL` and `useSVG` picked it up from a passive `useEffect`, which runs after the browser has painted. OSD raises `viewport-change` from its animation-frame update, after the springs advanced and before it draws the world, so a draw issued directly from the handler lands in the same frame.

- `Viewer`: remove the `viewport` state and `updateViewport`; the `viewport-change` handler (and the initial `getViewport()`) call two imperative setters instead. Pan/zoom no longer triggers any React re-render.
- `useWebGL`: drop the `viewport` parameter and its effect; add a stable `setGLViewport` that sets the viewport on both renderers and redraws synchronously if it changed. A viewport arriving during the async points-renderer init is re-applied before the first draw.
- `useSVG`: same treatment with `setSVGViewport`.
- The `rectEquals` dedup removed from `Viewer` is already done inside `WebGLRendererBase.setViewport` and `SVGController.setViewport`.
- Comments explain the mechanism and why the callbacks are memoized (review rounds).

# Screenshot of the fix

Not meaningful as a still image; the lag is only visible while dragging.

# Use of AI

Yes, help of Claude Code (Fable 5.1).

# Testing

- Instructions on how to test: `pnpm run dev`, open a project with points and/or shapes over an image, pan and zoom by dragging and wheel. Before: markers and shapes visibly trail the image by a frame and snap back when the movement stops. After: they move rigidly with the image. Draw a selection with the SVG tools while panning to check the SVG layer too.
- `tsc`, `eslint` and `prettier` pass on the touched packages.

# Further comments

The fix has been re-introduced once already. If the hooks ever grow a viewport parameter again, that is the regression.

### DIFF
--- a/packages/@tissuumaps-viewer/src/hooks/useWebGL.ts
+++ b/packages/@tissuumaps-viewer/src/hooks/useWebGL.ts
@@ -16,6 +16,14 @@ type GL = {
   shapesRenderer: WebGLShapesRenderer;
 };
 
+/**
+ * Clears the canvas and redraws both WebGL renderers
+ *
+ * Deliberately a module-level function of the GL object rather than a callback
+ * inside the hook: it keeps one identity for the lifetime of the module, so the
+ * memoized setters below cannot capture a stale copy of it and no hook has to
+ * list it as a dependency.
+ */
 function drawGL(gl: GL | null) {
   if (gl !== null) {
     gl.context.clear();
--- a/packages/@tissuumaps-viewer/src/hooks/useWebGL.ts
+++ b/packages/@tissuumaps-viewer/src/hooks/useWebGL.ts
@@ -68,9 +68,7 @@ export function useWebGL(
     }
   }
 
-  // called synchronously from the viewport-change handler in Viewer (see there);
-  // stable identity, as Viewer's effect depends on it and would otherwise
-  // recreate the whole overlay on every render
+  // called synchronously from the viewport-change handler in Viewer (see there)
   const setGLViewport = useCallback((viewport: Rect) => {
     viewportRef.current = viewport;
     if (glRef.current !== null) {
--- a/packages/@tissuumaps-viewer/src/hooks/useWebGL.ts
+++ b/packages/@tissuumaps-viewer/src/hooks/useWebGL.ts
@@ -132,7 +132,6 @@ export function useWebGL(
       let shapesRenderer: WebGLShapesRenderer;
       try {
         shapesRenderer = new WebGLShapesRenderer(context, {
-          viewport: viewportRef.current ?? undefined,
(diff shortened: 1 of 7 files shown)


packages/@tissuumaps-viewer/src/hooks/useWebGL.ts
```diff
@@ -16,9 +16,16 @@ type GL = {
   shapesRenderer: WebGLShapesRenderer;
 };
 
+function drawGL(gl: GL | null) {
```
jwindhager, line 27: Perhaps good to leave a comment here why drawGL is a module-level function?

packages/@tissuumaps-viewer/src/hooks/useWebGL.ts
jwindhager: For consistency with the WebGLPointsRenderer, can we defer setting the viewport here as well? If we do, also remove the viewport from the `options` (see above)

packages/@tissuumaps-viewer/src/hooks/useWebGL.ts
```diff
@@ -95,7 +105,6 @@ export function useWebGL(
           resolvePointsRendererInitPromise,
           rejectPointsRendererInitPromise,
           {
```
jwindhager, line 130: I think the WebGLPointsRenderer `options` still take an optional `viewport`, right? If this should not be used (defer setting viewport to after init, see below), we should maybe remove that...

jwindhager (changes requested):
LGTM, mostly, just a few changes requested.

Regarding:
> Container size still travels through React state; a resize lags one frame for the same reason. Left for a follow-up to keep this PR viewport-only.

I think it would be good to immediately bypass react state mgmt for containerSize as part of this PR, too (if it really is the same pattern --> to be confirmed).

cavenel:
Thanks @jwindhager! 
All comments should now be fixed:
 - added a comment for `drawGL`.
 - draw the overlay synchronously on container resize
 - set the initial viewport with setViewport, not in the constructor
 - Add `shapesRenderer.setViewport(viewportRef.current);`, same as the `pointsRenderer`

jwindhager (approved):

PR #176, opened 2026-09-07
Feat/table column input

cavenel:

Restores the column-picker behaviour of TissUUmaps 3, where the field was a free-text path input with autocomplete instead of a combobox.

# Type of change

Internal only: the `SimpleAsyncCombobox` component is removed, no project file or data source format changes.

# List of changes made

**Why.** The "Table column" fields used a base-ui `Combobox` in single-selection mode. That model fought the use case: opening the popup blanked the field, Enter or blur after typing discarded the text, and hierarchical suggestions (HDF5 `obs/…`) could not be descended into. Wide tables also rendered every column in the popup.

**`@tissuumaps/core`**
- New `TableColumnUtils` (with tests) holding the column query suggestion and resolution previously duplicated in `CSVTableData` and `ParquetTableData`. Suggestions match case-insensitively with the exact match listed first. A query resolves to the exactly named column, or to the single column differing only in case.
- `TableData.suggestColumnQueries` documents the group convention: a suggestion ending in `/` continues the query into a group instead of resolving to a column.

**`@tissuumaps/storage`**
- `CSVTableData` and `ParquetTableData` delegate to `TableColumnUtils` and now honour the abort signal.

**App**
- New `TableColumnInput` widget (`components/widgets/TableColumnInput`) on base-ui `Autocomplete`: the typed text is the value, suggestions fill it, Enter or blur commits it through `resolveColumnQuery`, unresolved text stays in the field marked invalid without changing the store, the clear button unsets the column, a pressed group suggestion reopens the popup on its children, and the list is capped at 100 entries with a "showing 100 of N" hint.
- `SimpleAsyncCombobox` deleted; the six consuming widgets pass `tableId`, `value` and `onValueChange` instead of five props plus a hook call.
- `useTableColumnSelector` moved next to its only consumer as `useTableColumnQueries`.

**Decisions taken, for reviewers**
- The cap is applied in the widget, not in the providers, so truncation is detected exactly and providers have one rule fewer.
- Groups are handled by letting base-ui close the popup and reopening it, not by cancelling the close: base-ui only resets the highlighted index on unmount, so cancelling would leave a stale highlight on the children.
- Group support ships before any merged provider produces groups. The HDF5 provider branch is the producer.
- Enter on typed text commits the text rather than the first suggestion (no `autoHighlight`).

# Screenshot of the fix

# Use of AI

Claude Code was used throughout: investigation of the old component and the TissUUmaps 3 behaviour, the implementation plan, the code and tests. All changes were reviewed by a human.

# Testing

- A rebuild is needed: the app consumes `@tissuumaps/core` and `@tissuumaps/storage` from `dist`, run `pnpm run build:packages` before `pnpm dev`.
- Automated: `pnpm test` (new `TableColumnUtils.test.ts`), `pnpm lint`, `pnpm exec tsc -p apps/tissuumaps/tsconfig.json --noEmit`.
- Manual, with a project containing a CSV or Parquet table, in any marker config widget or the "Group by" field:
  1. Click the field: the popup lists columns and the current text is kept and editable.
  2. Type part of a name in the wrong case: the list filters case-insensitively, exact match first.
  3. Enter with a valid full name, popup open or closed: the column is applied.
  4. Enter with an unknown name: the text stays, the field shows the invalid ring, the store is unchanged.
  5. Arrow down + Enter, and mouse click: the item is applied.
  6. Tab away with a valid name: applied once; with an unknown name: invalid, unchanged.
  7. Clear button: the column is unset. Escape when closed: the text is preserved.
  8. Table with more than 100 columns: the popup shows 100 and the "keep typing" hint.
  9. Group descent requires the HDF5 branch rebased on this PR: clicking `obs/` reopens the popup on its children.

# Further comments

A column literally named with a trailing `/` cannot be picked from the suggestion list under the group convention, but typing it and pressing Enter still resolves it. Documented on the contract; judged acceptable for a `string[]` contract.


apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+  const [invalid, setInvalid] = useState(false);
+  const [suggestions, setSuggestions] = useState<string[] | null>(null);
+  const [open, setOpen] = useState(false);
+
+  const [syncedValue, setSyncedValue] = useState(value);
+  if (value !== syncedValue) {
+    setSyncedValue(value);
+    setText(value ?? "");
+    setInvalid(false);
+  }
```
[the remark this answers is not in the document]
cavenel: I believe this is the correct way of doing it. Won't fix.

packages/@tissuumaps-core/src/utils/TableColumnUtils.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * Column query suggestion and resolution for tables with a flat list of columns
+ *
+ * A column query is the column name itself.
+ */
+export class TableColumnUtils {
```
jwindhager, line 6: I'd suggest to rename this to TableUtils (the functions already say "column")

packages/@tissuumaps-core/src/utils/TableColumnUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+   */
+  static suggestColumnQueries(
+    columns: string[],
+    currentQuery: string,
+  ): string[] {
+    const lowerCaseQuery = currentQuery.toLowerCase();
+    const matches: string[] = [];
+    const others: string[] = [];
+    for (const column of columns) {
+      if (column.toLowerCase().includes(lowerCaseQuery)) {
```
jwindhager: startsWith or includes?

packages/@tissuumaps-storage/src/parquet/ParquetTableData.ts
```diff
@@ -45,15 +46,15 @@ export class ParquetTableData implements TableData {
   }
 
   suggestColumnQueries(currentQuery: string): Promise<string[]> {
-    const filteredColumns = this._columns.filter((column) =>
-      column.includes(currentQuery),
+    return Promise.resolve(
+      TableColumnUtils.suggestColumnQueries(this._columns, currentQuery),
```
jwindhager, line 50: Does this mean that the parquet data loader loads all columns upfront? I don't know the format well - is this performant/acceptable?

packages/@tissuumaps-core/src/utils/TableColumnUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+        matches.push(column);
+      } else {
+        others.push(column);
+      }
+    }
+    const exactMatchIndex = matches.indexOf(currentQuery);
+    if (exactMatchIndex > 0) {
```
jwindhager: Why not do this in the loop already?

packages/@tissuumaps-core/src/utils/TableColumnUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+      matches.unshift(...matches.splice(exactMatchIndex, 1));
+    }
+    return [...matches, ...others];
+  }
+
+  /**
+   * Resolves a column query to a column name
+   *
+   * The query resolves to the column with the same name, or otherwise to the
+   * single column whose name differs only in case.
```
jwindhager, line 55: Can we assume case-insensitivity for all formats here? The previous strategy was to *suggest* columns case-insensitively, but users have to *select* columns case-sensitively (in case there are two columns that just differ in case).

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+export type TableColumnInputProps = {
+  tableId: string | null;
+  value: string | null;
+  onValueChange: (column: string | null) => void;
+  className?: string;
+};
+
+/** Maximum number of suggestions shown in the popup, which is not virtualized */
+const maxSuggestions = 100;
```
jwindhager: Can we make this an optional prop?

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+function isGroupQuery(query: string): boolean {
+  return query.endsWith("/");
```
jwindhager, line 21: Can we assume that `/` (slash) is the path separator for all hierarchical formats? Zarr v2 for example allowed arbitrary `dimension_separator` values, and Zarr v3 currently defines two official core chunk key encodings, both of which restrict separators to `/` (slash) or `.` (dot). In the previous design, we didn't consider group boundaries - it was up to the data provider to do that in a format-specific way in `suggestColumnQueries`.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+}
+
+function isMatch(suggestion: string, query: string): boolean {
+  return suggestion.toLowerCase().includes(query.toLowerCase());
```
jwindhager: This duplicates TableColumnUtils functionality (same concerns regarding case sensitivity apply)

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+}
+
+/**
+ * Loads the data of the table identified by `tableId` on demand
+ *
+ * @param tableId - The ID of the table to load
+ * @returns A callback yielding the table data, or `null` if `tableId` does not
+ * identify a table of the current project
+ */
+function useLoadTableData(tableId: string | null) {
```
jwindhager: Can we inline this into the component?

jwindhager (changes requested):
Thanks for this! I think it's definitely going into the right direction, replacing the old ComboBox. However, before I review the `SuggestionText` and `TableColumnInput` components in detail, I have a few functional suggestions/concerns.

packages/@tissuumaps-storage/src/parquet/ParquetTableData.ts
cavenel: All columns names (read from the Parquet footer when the file is opened) so I would say acceptable!

packages/@tissuumaps-core/src/utils/TableColumnUtils.ts
cavenel: I would keep as is, it's not a helper for tables, the file is only about column queries. I would follow the rest of Utils (ColorUtils, GeometryUtils, NumberUtils, ...) with the filename describing the utils in it.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: Done.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: Fixed. There is now one `TableColumnUtils.matchColumnQuery(column, query)` returning the match index or -1. It is used by core for ordering, by the suggestion list for the bold substring, and by the status message.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: The `/` is not a format thing. The query language is ours, so a Zarr provider with `.` separators would still return `/`-terminated queries. So Zarr is fine I think.

The main problem I see is if a csv column ends in "/", then we are stuck. But can be fixed easily by adding a `isgroup` that would be false for csv.

packages/@tissuumaps-core/src/utils/TableColumnUtils.ts
cavenel: Suggestion is case-insensitive, resolution only accepts an unambiguous match. So I think that follows the same idea!

packages/@tissuumaps-core/src/utils/TableColumnUtils.ts
cavenel: Done.

packages/@tissuumaps-core/src/utils/TableColumnUtils.ts
cavenel: includes, deliberately. In between a search and an autocomplete. Proven to be powerful for users!

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: Done, maxSuggestions is now an optional prop defaulting to 100.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
jwindhager: Ah, good one, so we have our own (very simple) query language! :)

jwindhager: I thought about this a bit further:
- In the original design (before this PR), the query format was a provider-only concern and opaque to the UI: a provider returned suggestions for a partial query and resolved a complete query to a column name (usually, query=column name in this case); nothing in the UI needed to interpret the strings.
- Specifically, the intended contract was: while typing/selecting, the UI would call `suggestedQueries = suggestColumnQueries(currentQuery)` to retrieve & update the list of suggestions, and `column = resolveColumnQuery(currentQuery)` to check whether to commit (`column !== null`) or keep searching (`column === null`). The old UI didn't need to distinguish groups from leaves, but it could've done so anytime by calling `resolveColumnQuery` on each suggestion returned by `suggestColumnQueries` (e.g. to visually distinguish groups from leaves in the list of suggestions).
- If one wanted to save the async `resolveColumnQuery` roundtrip to check whether a suggestion is a group or leaf, `suggestColumnQueries` could have additionally returned a `terminal: boolean` flag with each suggestion. Similarly, one could've added `label?: string` to the suggestions in case one doesn't want to show the full path prefix in the suggestions, or a `match?:  [startIndex, endIndex]` for highlighting matches.

Unless I'm missing something, none of these examples/features would require the introduction of a new custom query language, and I don't see why this PR would need to either? If that's the case, I would prefer to leave the format entirely to the data providers as with the previous design, for the following reasons:
- Avoid implementation mistakes in the data providers, which would otherwise need to translate between our query language and their DSL
- Give data providers the opportunity to support more powerful query languages such as regular expressions, XPath/JSONPath selectors, etc. For example, note that indexing and querying extensions are a frequently discussed topic not only in Zarr implementations (e.g. duckdb_zarr, zarr-datafusion, xarray-sql), but also on the Zarr spec level itself (e.g. spatial queries to find all points within a bounding box, based on [zarr-conventions/spatial](https://github.com/zarr-conventions/spatial) metadata)
- Remain transparent to the user: if, for example, a format always uses `.` (e.g. JSONPath, SQL schema, some Zarr v3) as path separator and the users are used to it, why should we expect them to switch?

cavenel: Yep I think your `terminal: boolean` etc. would solve it nicely. I will try that.

packages/@tissuumaps-core/src/storage/table.ts
```diff
@@ -12,16 +25,19 @@ import type { DataProvider, ItemsData } from "./base";
...
+   * @param currentQuery - The partial column query to autocomplete
    * @param options - Optional abort signal
-   * @returns A list of suggested column queries matching the current one
+   * @returns A list of suggested column queries
    */
   suggestColumnQueries(
     currentQuery: string,
     options?: { signal?: AbortSignal },
-  ): Promise<string[]>;
+  ): Promise<ColumnQuerySuggestion[]>;
```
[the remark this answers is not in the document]
cavenel: Yes, this is a breaking change, and it is intended. The commit is marked !. I updated the PR description to say so. We will not keep the old string[] contract next to the new one.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+          console.error("Failed to suggest column queries", error);
+        }
+      }
+    });
+    return () => abortController.abort();
+  }, [open, text, loadTableData, startSuggestTransition]);
+
+  const [isCommitPending, startCommitTransition] = useTransition();
+  const commitAbortControllerRef = useRef<AbortController | null>(null);
+  useEffect(() => () => commitAbortControllerRef.current?.abort(), []);
```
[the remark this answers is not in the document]
cavenel: A pending commit is now also aborted when value or the table changes.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+    if (suggestions === null) {
+      return isSuggestPending ? "Loading table..." : null;
+    }
+    if (suggestions.length === 0) {
+      return "No columns";
+    }
+    // matching suggestions are listed first, so the first one decides
+    if (TableColumnUtils.matchColumnQuery(suggestions[0]!.query, text) === -1) {
+      return `No matches for "${text}", showing all columns`;
+    }
```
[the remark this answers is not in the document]
cavenel: Fixed. When the list is capped, the message now says "showing the first 100 columns".

packages/@tissuumaps-storage/src/csv/CSVTableData.ts
```diff
@@ -43,16 +45,16 @@ export class CSVTableData implements TableData {
     return this._names;
   }
 
-  suggestColumnQueries(currentQuery: string): Promise<string[]> {
-    const filteredColumns = this._columns.filter((column) =>
-      column.includes(currentQuery),
+  suggestColumnQueries(currentQuery: string): Promise<ColumnQuerySuggestion[]> {
```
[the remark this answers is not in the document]
cavenel: Won't fix. These methods are synchronous, so there is nothing to cancel. The caller checks the signal after the call. The PR description was wrong about this, I updated it.

packages/@tissuumaps-storage/src/parquet/ParquetTableData.ts
```diff
@@ -44,16 +46,16 @@ export class ParquetTableData implements TableData {
     return this._names;
   }
 
-  suggestColumnQueries(currentQuery: string): Promise<string[]> {
-    const filteredColumns = this._columns.filter((column) =>
-      column.includes(currentQuery),
+  suggestColumnQueries(currentQuery: string): Promise<ColumnQuerySuggestion[]> {
```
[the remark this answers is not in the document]
cavenel: Same as in CSVTableData: the method is synchronous, so won't fix.

packages/@tissuumaps-core/src/utils/TableColumnUtils.ts
jwindhager: Now that we have TableUtils, do you still think we should keep a separate TableColumnUtils?

cavenel: Removed the Utils instead!

packages/@tissuumaps-core/src/storage/table.ts
```diff
@@ -3,6 +3,19 @@ import type { GenericArray } from "../types/arrays";
 import type { ProgressCallback } from "../types/callbacks";
 import type { DataProvider, ItemsData } from "./base";
 
+/** A column query suggested by `TableData.suggestColumnQueries` */
+export type ColumnQuerySuggestion = {
+  /** The suggested query, in the provider's own format */
+  query: string;
+  /**
```
jwindhager:
```suggestion

  /**
```

packages/@tissuumaps-core/src/storage/table.ts
```diff
@@ -3,6 +3,19 @@ import type { GenericArray } from "../types/arrays";
 import type { ProgressCallback } from "../types/callbacks";
 import type { DataProvider, ItemsData } from "./base";
 
+/** A column query suggested by `TableData.suggestColumnQueries` */
+export type ColumnQuerySuggestion = {
```
jwindhager: I'd name this `TableColumnQuerySuggestion` or inline it

packages/@tissuumaps-core/src/storage/table.ts
```diff
@@ -3,6 +3,19 @@ import type { GenericArray } from "../types/arrays";
...
+  /** The suggested query, in the provider's own format */
+  query: string;
+  /**
+   * Whether the query resolves to a column
+   *
+   * A non-terminal suggestion continues the query, e.g. into a group of
+   * columns.
+   */
```
jwindhager:
```suggestion
   * Whether the query directly resolves to a column
   *
   * A terminal suggestion is expected to resolve to an exact column name;
   * a non-terminal suggestion continues the query, e.g. into a group of
   * columns.
   */
```

packages/@tissuumaps-core/src/storage/table.ts
```diff
@@ -3,6 +3,19 @@ import type { GenericArray } from "../types/arrays";
...
+  terminal: boolean;
```
jwindhager: Do we perhaps want to invert this to an optional `group?: boolean`? Then non-hierarchical providers would only need to return `{ query }`.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+  query: string;
+};
+
+function SuggestionText({ suggestion, query }: SuggestionTextProps) {
+  const index = query !== "" ? findQuery(suggestion, query) : -1;
+  if (index === -1) {
+    return <span className="truncate">{suggestion}</span>;
+  }
+  return (
+    <span className="truncate">
```
jwindhager: Perhaps good to truncate in the beginning rather than in the end? What do you think?

```suggestion
    <span className="truncate [direction:rtl] text-left">
```

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+}: TableColumnInputProps) {
+  const table = useProjectStore(
+    (state) => state.tables.find((table) => table.id === tableId) ?? null,
+  );
+  const loadTable = useTableDataLoader();
+  const loadTableData = useCallback(
+    async (signal?: AbortSignal) =>
+      table !== null ? await loadTable(table, { signal }) : null,
+    [table, loadTable],
+  );
```
jwindhager: This pattern (loadTableData = useTableDataLoader + table) sits "in between" `useTableDataLoader` and `useTableData` and is/may be reused in several places - could we add a new `useLazyData.ts` containing `useLazyXData(xId: string | null): (signal?: AbortSignal) => Promise<XData | null>` hooks for this?

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+    ColumnQuerySuggestion[] | null
+  >(null);
+  const [open, setOpen] = useState(false);
+
+  const [syncedValue, setSyncedValue] = useState(value);
+  if (value !== syncedValue) {
+    setSyncedValue(value);
+    setText(value ?? "");
+    setInvalid(false);
+  }
```
jwindhager: I'm not sure I understand this block - it smells a bit like this can/should be done differently?

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+    return () => abortController.abort();
+  }, [open, text, loadTableData, startSuggestTransition]);
+
+  const [isCommitPending, startCommitTransition] = useTransition();
+  const commitAbortControllerRef = useRef<AbortController | null>(null);
+  useEffect(
+    () => () => commitAbortControllerRef.current?.abort(),
+    [value, loadTableData],
+  );
+  function commit(query: string) {
```
jwindhager: Would it be cleaner to store the pending commit as state and resolves it in an effect, like the suggest path already does? I think the ref and the effect that exists only to abort would then both go away.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+      itemToStringValue={(suggestion) => suggestion.query}
+      openOnInputClick
+      open={open}
+      onOpenChange={handleOpenChange}
+      onOpenChangeComplete={handleOpenChangeComplete}
+      onItemHighlighted={(suggestion) => {
+        highlightedSuggestionRef.current = suggestion;
+      }}
+    >
+      <div className={cn("relative w-full", className)}>
```
jwindhager: Why do we need that extra div?

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
```diff
@@ -0,0 +1,292 @@
...
+            <ChevronDownIcon className="size-4" />
+          </Autocomplete.Trigger>
+        </div>
+      </div>
+      <Autocomplete.Portal>
+        <Autocomplete.Positioner
+          className="isolate z-50 outline-none"
+          sideOffset={4}
+        >
+          <Autocomplete.Popup className="box-border w-(--anchor-width) max-h-[min(var(--available-height),23rem)] max-w-(--available-width) origin-(--transform-origin) overflow-y-auto overscroll-contain rounded-md border bg-popover py-1 text-popover-foreground shadow-md transition-[transform,scale,opacity] data-ending-style:transition-none data-starting-style:scale-95 data-starting-style:opacity-0">
```
jwindhager: There is lots of styling in here, because we use the base-ui Autocomplete directly. shadcn/ui doesn't have a sibling for this (the official "alternative" for this is Combobox, which doesn't serve our use case well), so this makes sense. However, I would consider factoring out the styling into our own `autocomplete.tsx` in `components/common`, similar to e.g. `accordion.tsx`.

jwindhager (commented):
Looks mostly good to me

Some minor comments and architectural suggestions

packages/@tissuumaps-core/src/storage/table.ts
cavenel: Done!

packages/@tissuumaps-core/src/storage/table.ts
cavenel: Renamed to TableColumnQuerySuggestion and kept it as a named type, since the widget and the hierarchical provider both refer to it.

packages/@tissuumaps-core/src/storage/table.ts
cavenel: Fixed

packages/@tissuumaps-core/src/storage/table.ts
cavenel: Good idea, done

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: Agreed, done (a bit differently, so that punctuation doesn't move to start, especially /)

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: Added hooks/useLazyData.ts with useLazy{Image,Labels,Points,Shapes,Table}Data, mirroring useData.ts. The input now uses useLazyTableData(tableId).

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: Renamed to prevValue (to fit other examples in the code) and added the link.

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: Yes, that's cleaner. Done!

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: It was the positioning box for the absolutely placed clear and show-columns buttons. Replaced with base-ui's Autocomplete.InputGroup rendered as our InputGroup, with the buttons in an InputGroupAddon, like the seed field in ColorConfigWidget. Cleaner!

apps/tissuumaps/src/components/widgets/TableColumnInput/index.tsx
cavenel: Done in `components/common/autocomplete.tsx`.

jwindhager (approved):
🎉

PR #177, opened 2026-09-07
feat: basic labels rendering via OpenSeadragon data transfers

jwindhager:

Jira: [TMAP-75 Basic labels rendering](https://scilifelab.atlassian.net/browse/TMAP-75)

# Type of change

- [ ] Other

# List of changes made

**Data transfers (`@tissuumaps/render`)**

- `OpenSeadragonContext`: tiles whose pixels carry values (label IDs, channel intensities) are recolored in the `tile-invalidated` handler by a `DataTransfer` (`getData` + `transfer`), kept per tile source and compared by identity. Replaces per-tile tinting (`setTiledImageTint` removed). Colors are written as packed `0xAABBGGRR` through a `Uint32Array` view of a fresh canvas sized from the tile's source bounds, with a byte swap on big-endian hosts.
- `OpenSeadragonRendererBase`: `synchronize` moved into the base and takes a typed sync context (`OpenSeadragonSyncContext`, with `loadObject`) instead of positional loaders. Subclasses hook in via `getTileSources`, `usesAdditiveBlending`, `getTiledImageOpacity` and the new `getTiledImageDataTransfer`, `resolveObject` (called per object once its data has loaded, the only hook that sees the sync context) and `retainObjects` (drops per-object state of objects no longer displayed); former protected internals are private.
- `OpenSeadragonLabelsRenderer`: implemented (was a placeholder). Resolves per-label color, visibility and opacity with the shared resolvers, folds them into an ID → RGBA lookup table and applies it as a data transfer. Label `0` is transparent, unlisted IDs use the defaults. The transfer is cached per object until its data or label configuration changes.
- `OpenSeadragonImageRenderer`: multi-channel images are recolored per channel from color and contrast limits (image `Channel` settings, falling back to `ImageData` getters), cached per channel. Channels without a color use a default color derived from the channel index (`RenderUtils.getDefaultChannelColor`), channels without contrast limits use the value range of their data type (`RenderUtils.getDataTypeRange`); channels without channel data are drawn as-is.
- Resolvers moved from `webgl/resolvers/` to `resolvers/`, as they are now shared by the OpenSeadragon renderers.

**Interfaces (`@tissuumaps/core`)**

- `LabelsData` reworked for the OpenSeadragon 6 data pipeline: `getTileSource()` and `getData(event)` replace the level/tile getters and `loadTile`.
- `ImageData`: new optional `getChannelData` and `getChannelContrastLimits`; `getChannelName` no longer returns `undefined`. The interface docs now spell out the channel data contract: `getChannelData` provides single samples per pixel that are contrast-stretched and colorized by the renderer, and providers with color information for only some channels should return default colors for the remaining ones themselves, so that metadata colors and the index-based fallback do not mix within one image.
- `Channel.contrastLimits` added. New `RenderUtils` with the defaults the renderer applies: `getDefaultChannelColor` assigns default channel colors by index (red, green, blue, yellow, cyan, magenta, then HSB-derived hues; wraps after 100 channels), resembling QuPath's defaults; `getDataTypeRange` returns the default contrast limits per array type (full integer range, `[0, 1]` for floats, `[0, 255]` for plain arrays). `ColorUtils.fromHSB` added (public).
- Packed color layout switched to `0xAABBGGRR` (red in the lowest byte) so packed colors match canvas `ImageData` byte order: `ColorUtils.packColor`, `points.vert`/`shapes.frag` `unpackColor`, and `WebGLRendererBase.foldRGBA` (renamed from `packAlpha`). The alpha folding itself is now shared as `ColorUtils.packRGBA`, used by the WebGL renderers and the labels renderer.
- `types/render.ts` split into `types/openseadragon.ts` (now also home of `TileSourceConfig`/`CustomTileSource`) and `types/webgl.ts`.

**App / viewer**

- `ImageDataWrapper` forwards optional channel getters only when the wrapped data provides them; `LabelsDataWrapper` follows the new interface.
- `useOpenSeadragon` passes sync context objects to the renderers.

# Screenshot of the fix

n/a

# Use of AI

Claude Code was used throughout development: pair-programming parts of the implementation, reviewing the changes, and drafting commit messages and this PR description. All changes were reviewed by the author.

# Testing

- No rebuild or configuration changes needed.
- Rendering of regular (single-channel / RGB) images, points and shapes should be unchanged; verify colors in the WebGL renderers are correct after the byte-order switch.
- Labels and channel recoloring cannot be exercised end-to-end yet, see below.

# Further comments

- Draft: no `LabelsData` provider exists in the repo yet, and no `ImageData` provider implements `getChannelData`, so labels rendering and channel recoloring are wired in the renderers but inactive until providers are added. Until then, `Channel.color` on multi-channel images has no visible effect (previously applied as a tint).
- Public interface changes (`LabelsData`, renderer `synchronize` signatures, `packColor` byte order, `types/render` split) are breaking for package consumers.


packages/@tissuumaps-core/src/model/constants.ts
```diff
@@ -13,6 +13,13 @@ export const identityTransform = {
   translation: { x: 0, y: 0 },
 } as const satisfies SimilarityTransform;
 
+/** Default image channel color */
```
cavenel: `defaultChannelColor` is white and index independent. A channel with contrast limits but no file color renders white, so a 4-plex with no omero or OME colors is four white channels added together into one white blob, that's unfortunate.

Should each provide come with its own list of default colors? (like avivator: https://github.com/hms-dbmi/viv/blob/ec590278f8f5591c64200be9ce189aa5c12215fc/sites/avivator/src/constants.js#L19)
Or should it be in the model, and then be index aware? Suggest a small ordered palette plus a lookup that wraps on the channel index.

packages/@tissuumaps-core/src/model/image.ts
```diff
@@ -21,30 +21,43 @@ export const imageDefaults = {} as const satisfies Partial<RawImage>;
...
+   * Overrides the channel name provided by the image data provider, if any.
    */
   name?: string;
 
   /**
    * Channel visibility
    *
+   * Overrides the channel visibility provided by the image data provider, if any.
+   *
    * @defaultValue `true`
```
cavenel: `@defaultValue "true"` contradicts the line above it. Absent now means "whatever the provider reports, otherwise true".

Same on line 44 for `@defaultValue "1"`, and in the `channels` doc on `RawImage`.

```suggestion
   * @defaultValue The visibility reported by the image data provider, or `true`
```

packages/@tissuumaps-core/src/storage/image.ts
```diff
@@ -46,31 +53,31 @@ export interface ImageData extends Data {
...
-  getChannelName?: (c: number) => string | undefined;
+  /**
+   * Extracts the raw image data of a specific channel from a tile invalidation event
+   *
+   * @param c - The channel index (0-based)
+   * @param event - The tile invalidation event
+   * @returns The channel's values for the invalidated tile, one per tile pixel
+   * in row-major order
+   */
+  getChannelData?: (
```
cavenel, line 83: Do we still need the `c` parameter? The event's tile already belongs to that channel's tile source.

I would say drop it, or say why it is there. (This also touches `OpenSeadragonImageRenderer` and the wrapper)

packages/@tissuumaps-core/src/storage/image.ts
```diff
@@ -46,31 +53,31 @@ export interface ImageData extends Data {
...
+    c: number,
+    event: OpenSeadragon.TileInvalidatedEvent,
+  ) => Promise<NumericArray>;
+
+  /** Returns the name of a specific channel */
+  getChannelName?: (c: number) => string;
```
cavenel: `getChannelName` returns `string`. Many files carry no channel names, so a provider has no honest value and has to invent a UI label like `Channel 0`. That decides presentation in the storage layer.

Return `string | undefined` and build the fallback label in the UI. `ImageDataWrapper` derives its type from this, so it follows automatically.

```suggestion
  /** Returns the name of a specific channel, or undefined if the data has none */
  getChannelName?: (c: number) => string | undefined;
```

packages/@tissuumaps-core/src/storage/labels.ts
```diff
@@ -30,56 +35,21 @@ export interface LabelsDataProvider<
...
   /**
-   * Loads a single tile as an unsigned integer array
+   * Extracts the raw label image data from a tile invalidation event
    *
-   * @param level - Pyramid level
-   * @param x - Tile column index
-   * @param y - Tile row index
-   * @param options - Optional abort signal and progress callback
+   * @param event - The tile invalidation event
+   * @returns The label image data as an unsigned integer array
```
cavenel: The doc says "unsigned integer array", but the type is `number[] | UintArray`. `getChannelData` returns `NumericArray` for the same thing. Worth aligning the two types as well.

```suggestion
   * @returns The label image data, as a plain array or an unsigned integer array
```

packages/@tissuumaps-core/src/storage/labels.ts
```diff
@@ -30,56 +35,21 @@ export interface LabelsDataProvider<
...
+   * @throws Error if the event does not contain label image data
    */
-  loadTile(
-    level: number,
-    x: number,
-    y: number,
-    options?: { signal?: AbortSignal; onProgress?: ProgressCallback },
-  ): Promise<UintArray>;
+  getData(
```
cavenel, line 56: Should Core depend on OpenSeadragon? (Same for getChannelData at storage/image.ts:64). I see two problems here:
- Every provider has to import an OSD event type.
- Core does not re-export it, so LabelsDataWrapper:35 writes Parameters<LabelsData["getData"]>[0]. Other wrappers spell their types out.

Not sure what the best solution is though...
- Pass a plain object with the tile geometry and the decoded data. No OSD type, no second fetch.
- Pass coordinates and a size. Drops the dependency, but the provider refetches a tile OSD already has. This is the old loadTile.
- Keep the event, re-export the type from types/openseadragon.ts, and say in the PR description that it is deliberate.

packages/@tissuumaps-core/src/utils/ColorUtils.ts
```diff
@@ -36,13 +37,38 @@ export class ColorUtils {
...
+   * Combines the lower 24 bits of `color` (see {@link packColor}) with
+   * `opacity` as alpha if `visibility` is non-zero, and with zero alpha
+   * otherwise, into a packed 32-bit RGBA color, `0xAABBGGRR`.
+   *
+   * @param color - The packed color, `0xBBGGRR`; any higher bits are discarded
+   * @param visibility - The visibility, `0` for invisible
+   * @param opacity - The opacity, in the range [0, 255]
+   * @returns The packed RGBA color, `0xAABBGGRR`, as an unsigned 32-bit integer
+   */
+  static packRGBA(color: number, visibility: number, opacity: number): number {
```
cavenel, line 66: Rename to `foldAlpha` maybe? (Closer to what it is really doing, as color is already packed from components)

`@param opacity` says `[0, 255]`, but nothing clamps. `packRGBA(c, 1, 256)` overflows the shift and returns alpha 0, so fully transparent!! Probably not what we would expect if pushing the opacity too much.

`@param visibility` says `0` means invisible, but the code tests `> 0`, so negatives are invisible too.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
```diff
@@ -77,30 +97,9 @@ export class OpenSeadragonContext {
...
-          // multiply with color
-          ctx.globalCompositeOperation = "multiply";
-          ctx.fillStyle = `rgba(${r}, ${g}, ${b}, 1)`;
-          ctx.fillRect(0, 0, width, height);
-          ctx.restore(); // pop context state
-          await event.setData(ctx, "context2d");
-        }
-      }
-    });
+    this.viewer.addHandler("tile-invalidated", (event) =>
```
cavenel, line 110: No `catch` on the handler (`_transferData` can reject from `_getTileSize`, or from `getData` when the wrapper throws `Data has been destroyed` after cache eviction.)
Wrap the body of `_transferData`, `console.error`, return. **A failed recolor should just leave the tile as it was**.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
```diff
@@ -350,54 +387,45 @@ export class OpenSeadragonContext {
...
-   * The tint is remembered per tile source, until the tile source is
-   * garbage-collected, because that is what OpenSeadragon keys its tile caches
-   * by: tiles that share their original data share their tinted data, too.
-   * Tiled images that share a tile source therefore also share a tint, with the
-   * last one set winning - including the tiled images of the navigator, which
-   * mirror those of this viewer and are tinted along with them.
+   * The data transfer is remembered per tile source, until the tile source is
+   * garbage-collected. Tiled images that share a tile source therefore also
+   * share a data transfer, with the last one set winning - including the tiled
+   * images of the navigator, which mirror those of this viewer and are
```
cavenel, line 412: Navigator tiles that are already loaded keep stale colors after a transfer changes. Each `Viewer` owns its `TileCache`, and `TiledImage.requestInvalidate` walks only its own (`openseadragon.js:29209`). New navigator tiles are fine, since their load-time invalidation is raised on the main viewer.

`Viewer.requestInvalidate` fixes it but invalidates every tile of every tiled image, which is expensive at the per-pixel cost of the labels transfer (we DON'T WANT THAT).

So: accept the stale navigator and correct this comment, which currently claims per-source keying "also covers the navigator", or invalidate wider and accept the cost (no).

(I haven't tested, so that probably needs more testing to be 100% sure)

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
```diff
@@ -350,54 +387,45 @@ export class OpenSeadragonContext {
...
-    if (
-      (oldTint === undefined && tint !== undefined) ||
-      (oldTint !== undefined && tint === undefined) ||
-      (oldTint !== undefined &&
-        tint !== undefined &&
-        !ColorUtils.colorsEqual(oldTint, tint))
-    ) {
-      if (tint !== undefined) {
-        this._tileSourceTints.set(tiledImage.source, tint);
+    const oldDataTransfer = this._tileSourceDataTransfers.get(
```
cavenel: Transfers are keyed by `tiledImage.source` but resolved per object id. `instantiateTileSourceClass` returns the same instance for an already built `TileSource`, and `DataCache` shares one data instance per normalized data source.

So two labels objects on the same URL with different colors flip this entry against each other every `synchronize`, and each flip fires a full invalidate.

Key the WeakMap by `tiledImage`.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
```diff
@@ -591,4 +581,84 @@ export class OpenSeadragonContext {
...
+    const dataTransfer = this._tileSourceDataTransfers.get(tiledImage.source);
+    if (dataTransfer === undefined) {
+      return;
+    }
+    const data = await dataTransfer.getData(event);
+    const { width, height } = OpenSeadragonContext._getTileSize(
+      event.tile,
+      data.length,
+    );
+    const canvas = document.createElement("canvas");
```
cavenel, line 668: One new canvas and one new `ImageData` per converted tile, on a hot path. A slider change reconverts every cached tile, so this arrives as a burst of garbage.

Reuse one module level canvas and buffer, resized as needed?

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
```diff
@@ -591,4 +581,84 @@ export class OpenSeadragonContext {
...
+   * actual tile size. The width is therefore taken from the bounds, rounded
+   * either way, and the height from the number of pixel values, which has to
+   * come out as an integer within one pixel of the bounds.
+   *
+   * @param tile - The tile whose size to determine
+   * @param n - The number of pixel values of the tile
+   * @returns The width and height of the tile, in source pixels
+   * @throws Error if no such size exists
+   */
+  private static _getTileSize(tile: OpenSeadragon.Tile, n: number): Dims {
```
cavenel: _getTileSize cannot reliably recover the tile shape from n / w, and gets it wrong silently.

So just remove `_getTileSize ` and make `getData` returns the shape.
Both drawers crop to `sourceBounds` themselves, so a full tile size canvas is fine, and providers no longer need to crop their rasters.

Proposed fix for getData:
```ts
getData(
  event: OpenSeadragon.TileInvalidatedEvent,
): Promise<{ values: number[] | UintArray; width: number; height: number }>;
```

Same for `getChannelData`.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonImageRenderer.ts
```diff
@@ -172,28 +184,139 @@ export class OpenSeadragonImageRenderer extends OpenSeadragonRendererBase<
...
+        getData: (event) => data.getChannelData!(index, event),
+        transfer: (values, buffer) => {
+          const [vmin, vmax] =
+            state.contrastLimits ??
+            OpenSeadragonImageRenderer._getDataTypeRange(values);
+          const scale = vmax > vmin ? 1 / (vmax - vmin) : 0;
+          for (let i = 0; i < values.length; i++) {
+            const v = MathUtils.clamp((values[i]! - vmin) * scale, 0, 1);
+            const color =
+              (Math.round(channelB * v) << 16) |
```
cavenel: This re-implements `ColorUtils.packColor` inline, depending on the `0xBBGGRR` layout without referring to it. It also rounds where `packColor` truncates, so the same `Color` packs differently on the image and labels paths.

Build a 256 entry `Uint32Array` ramp when the transfer is created, using `ColorUtils.packColor`. The hot loop becomes one lookup instead of three rounds and four bit operations.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonImageRenderer.ts
```diff
@@ -172,28 +184,139 @@ export class OpenSeadragonImageRenderer extends OpenSeadragonRendererBase<
...
+      return [-32768, 32767];
+    }
+    if (values instanceof Int32Array) {
+      return [-2147483648, 2147483647];
+    }
+    if (Array.isArray(values)) {
+      return [0, 255];
     }
-    return opacity;
+    return [0, 1];
```
cavenel: `return [0, 1]` is a catch-all, not a float branch. Float TIFF and Zarr data are rarely normalized, so a float channel with no limits saturates everywhere. `Uint8ClampedArray` also lands here, since it is not a `Uint8Array` subclass, and renders an 8-bit channel as a binary mask.

Test the float types explicitly and warn on anything unrecognised. At minimum, comment that float data is assumed normalised and providers are expected to supply limits.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonLabelsRenderer.ts
```diff
@@ -1,78 +1,263 @@
...
+ * kept (see {@link resolveObject}) until its data or one of the configurations
+ * it was resolved from changes, so that tiles are only recolored when needed.
  */
 export class OpenSeadragonLabelsRenderer extends OpenSeadragonRendererBase<
   Labels,
-  LabelsData
+  LabelsData,
+  OpenSeadragonLabelsSyncContext
 > {
+  private static readonly _defaultPixelValue = ColorUtils.packRGBA(
```
cavenel, line 64: `_defaultPixelValue` resolves to `0xffffffff`, opaque white. A label in the mask but missing from `getIds()` paints solid white over the image below.

Wouldn't transparent be safer and more what a user would expect?

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonLabelsRenderer.ts
```diff
@@ -1,78 +1,263 @@
...
+    data: LabelsData,
+    context: OpenSeadragonLabelsSyncContext,
+    options?: { signal?: AbortSignal },
   ): Promise<void> {
-    // TODO implement labels rendering
-    return Promise.resolve();
+    const renderedLabels = this._renderedLabels.get(labels.id);
+    if (
+      renderedLabels === undefined ||
+      renderedLabels.data !== data ||
```
cavenel, line 121: This compares `data` and the three label configs. It ignores `context.colorMaps`, `visibilityMaps` and `opacityMaps`, which `_resolveDataTransfer` reads and which do trigger a `synchronize`. So editing a color map does not recolor the labels.

Add the three maps to the same `deepEqual` check. They are already in the sync context.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonLabelsRenderer.ts
```diff
@@ -1,78 +1,263 @@
...
+      ),
+      OpacityResolver.resolveOpacities(
+        labelIds,
+        labels.labelOpacity,
+        context.opacityMaps,
+        defaultLabelOpacity,
+        { signal, loadTable },
+      ),
+    ]);
+    const labelPixelValues = new Map<number, number>();
```
cavenel, line 247: One `Map.get` per tile pixel, on the main thread. `requestInvalidate` covers every loaded tile, not just visible ones. That is pretty huge...

No easy fix here. You could build a `Uint32Array(maxId + 1)` prefilled with `_defaultPixelValue`, index 0 set to 0. That also drops the per-pixel `labelId === 0` branch. Keep the `Map` only when `maxId` is too big to allocate?

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonRendererBase.ts
```diff
@@ -14,38 +13,60 @@ import {
...
 
+/**
+ * The inputs of one {@link OpenSeadragonRendererBase.synchronize} call
+ *
+ * An immutable snapshot of the model state and loaders that a renderer needs,
+ * which subclasses extend with whatever else their objects resolve their
+ * appearance from. It carries inputs only; a renderer that derives state from
+ * an object does so in {@link OpenSeadragonRendererBase.resolveObject}.
+ */
+export type OpenSeadragonSyncContext<
```
cavenel: `OpenSeadragonSyncContext` spends two generics on a type with one field, and `OpenSeadragonImageSyncContext` instantiates it while adding nothing.

Delete it. Write the two contexts as plain types and constrain `TSyncContext` inline in the base class.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonRendererBase.ts
```diff
@@ -76,24 +97,78 @@ export abstract class OpenSeadragonRendererBase<
...
-    this._extraBounds = bounds;
+  async synchronize(
+    layers: Layer[],
+    objects: TObject[],
+    context: TSyncContext,
+    options?: { signal?: AbortSignal },
+  ): Promise<void> {
+    const { signal } = options ?? {};
+    signal?.throwIfAborted();
+    this.retainObjects(objects);
```
cavenel: `retainObjects` is documented as getting "the objects about to be displayed". It gets everything passed to `synchronize`, and `_loadObjects` filters by layer afterwards. Leftover entries keep a `LabelsData` and its LUT alive.

Call it after `_loadObjects`, keyed on the refs that survived.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonRendererBase.ts
```diff
@@ -154,25 +242,150 @@ export abstract class OpenSeadragonRendererBase<
...
+   */
+  protected retainObjects(
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _objects: TObject[],
+  ): void {}
+
+  /**
+   * Resolves what a renderer derives from an object, once its data has loaded
+   *
+   * Called by {@link synchronize} for every object, concurrently, right after
```
cavenel: `_loadObjects` calls this, and only for objects that pass the layer filter.

```suggestion
   * Called by {@link _loadObjects} for every object on one of the given layers,
   * concurrently, right after
```

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonRendererBase.ts
```diff
@@ -154,25 +242,150 @@ export abstract class OpenSeadragonRendererBase<
...
+
+  /**
+   * Concurrently loads and resolves all objects assigned to the specified layers
+   *
+   * Each object's data is loaded with the context's `loadObject`, and the object
+   * is then resolved (see {@link resolveObject}). The returned references are
+   * ordered by layer and then by object, which determines the order of the
+   * corresponding tiled images in the world. Objects whose data failed to load,
+   * or that could not be resolved, are logged and skipped; the logged error
+   * names the resolution as the cause, so that it is not mistaken for a failed
```
cavenel: The message actually printed is `Failed to load object with ID '...'`. The cause only shows if you expand it, so the distinction this doc promises is not visible.

Either change the message for a resolve failure, or drop the claim.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonRendererBase.ts
```diff
@@ -182,20 +395,29 @@ export abstract class OpenSeadragonRendererBase<
...
+        const newRefPromise = context
+          .loadObject(currentObject, { signal })
+          .then(async (data) => {
+            try {
+              await this.resolveObject(currentObject, data, context, {
+                signal,
+              });
+            } catch (error) {
+              if (!signal?.aborted) {
+                throw new Error("Failed to resolve object", { cause: error });
```
cavenel: A `resolveObject` rejection is handled like a failed data load, so the object is dropped and its tiled image removed. A label table that fails to load makes the whole layer vanish. Line 206 of the labels renderer handles a missing table by warning and falling back to defaults. Same cause, opposite result.

Keep the object. Log it, then render with no data transfer or with the default LUT.

cavenel (changes requested):
Very nice PR! Happy to discuss any of the comments, I couldn't always put a simple fix directly as some touch different files.

packages/@tissuumaps-core/src/storage/image.ts
jwindhager: I'd say we want to keep it for consistency with the other `getChannelX` methods. Also, future renderers may need the channel index, which cannot be extracted from the event (I think). Finally, it is somewhat of an architectural guard against calling this for single-channel images. Will add this to docs.

packages/@tissuumaps-core/src/storage/labels.ts
jwindhager: Unsigned integer array is correct for labels (`number[]` is expected to only hold unsigned integers in this case). `getChannelData` is for images and can therefore also return float/signed --> different, aligning those would be wrong.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
jwindhager: Do you mean you would wrap the _transferData call in a try/catch? Or wrap the whole body of _transferData (i.e., in _transferData)?

packages/@tissuumaps-core/src/utils/ColorUtils.ts
jwindhager: In my interpretation, "fold" would mean "apply visibility & opacity to the existing color value, see for example `WebGLRendererBase.foldRGBA`. Here, since this function operates on a single immutable number, we return a new value, which is why I named it "pack".

This function runs on individual values and needs to be fast. It therefore deliberately assumes correct ranges and deliberately doesn't clamp. The parameter ranges are documented as @param.

Fixed the visibility docs inaccuracy.

packages/@tissuumaps-core/src/storage/labels.ts
jwindhager: There are two levels to this question: the OpenSeadragon dependency in package.json, and being *architecturally* agnostic to OpenSeadragon in core.

Regarding the dependency in package.json, core depends on OpenSeadragon *for types only*, and OpenSeadragon is therefore just a dev/peer dependency (same as @jsonforms/core). Eventually, we want to be completely renderer-agnostic in core's dependencies, but currently we cannot (e.g. because OpenSeadragonOptions in the Project model). Data providers can follow a similar dev/peer dependency-only strategy for TS types during development.

Alternatively, to avoid the package.json dependency in data providers, instead of a full `getData(event): UnsignedArray`, we could ask data providers for `getDataType(): string`, i.e. for the `type` parameter of `TileInvalidatedEvent.getData()`, but this would not solve the architectural dependency problem.

I'd say we leave it for now, and revisit once we introduce modular rendering backends (--> architectural independence).

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
jwindhager: Fixed differently: https://github.com/TissUUmaps/TissUUmaps4/pull/177/commits/5cf1f0e8058f2b3e6155096aab2e2fd86565e738. Good?

cavenel: Yes way better nice job!

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
cavenel: I would go for the smaller change:
```
this.viewer.addHandler("tile-invalidated", (event) =>
  this._transferData(event).catch((error: unknown) => {
    console.error("Failed to transfer tile data", error);
  }),
);
```

packages/@tissuumaps-core/src/storage/labels.ts
cavenel: > I'd say we leave it for now, and revisit once we introduce modular rendering backends (--> architectural independence).

Sure, it works as is so no issue and can be fixed later in the modular backend refactoring!

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
jwindhager: Done in 9221a6d for the buffer half: the colors now go into a per-viewer `ArrayBuffer` that only grows, and each tile only creates typed-array views onto it (the `Uint32Array` for the transfer, the `Uint8ClampedArray` for the `ImageData`). Nothing awaits between filling the buffer and `putImageData`, which copies the pixels, so sharing is safe.

The canvas still has to be created per tile, though. OpenSeadragon keeps the value passed to `setData` by reference as the tile's cache record (`workingCache.addTile(tile, value, type)`, later injected as the main cache), and its `context2d` destroy hook zeros the canvas on eviction. A shared canvas would be overwritten by the next converted tile.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonImageRenderer.ts
jwindhager: I'd argue that it is common in the bioimaging domain to assume [0, 1] for float if no range is provided, and that this is already documented in the TSDoc comment. Uint8ClampedArray is currently not supported (see NumericArray). What would you like to change here?

cavenel: > I'd argue that it is common in the bioimaging domain to assume [0, 1] for float if no range is provided

True, and it is documented indeed! I was thinking of things like the output of a DeepLearning model having unbounded floats, but that's probably not our main target.
Ok to leave it as is!

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonLabelsRenderer.ts
jwindhager: Those are currently @todo in OpenSeadragonLabelsRenderer, WebGLPointsRenderer, WebGLShapesRenderer. Using deepEqual would work, but is costly for large maps. Perhaps better to check by identity (expecting any map update to to update its identity). We'll need to see how we implement this in the GUI (--> editable items table) before implementing here.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonLabelsRenderer.ts
jwindhager: I was thinking of using a LUT instead of a Map as you suggest, but that would assume contiguous labels. And I didn't want to prematurely optimize by branching depending on maxId (your suggestion). Lets first see how slow it is!

packages/@tissuumaps-core/src/model/constants.ts
jwindhager: I think we have the following options for channels without a user-configured/data-provided color:
1. Assume grayscale, as currently implemented.
2. Apply default colors from an application-global palette, as proposed.

I think option 1 is defensible, but perhaps not the most intuitive. In general, I would be careful with default colors, because we need to consider cases where color information is available for only a subset of images/channels. I think it could be confusing if channel 0 from image X is red because of file metadata, whereas X:1 or Y:0 is red because of configuration/data-agnostic defaults. Not sure how to address this.

Note: If providers really want to apply default colors, they could readily do so in `getChannelColor` (although I'd be hesitant to do so, for reasons already explained).

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonLabelsRenderer.ts
jwindhager: Currently, _defaultPixelValue (opaque white) is only applied in two cases:
- _resolveDataTransfer throws (i.e., table/column loading fails)
- label IDs not included in the label IDs returned by getIds()

Both cases are error states (getData should never return values that are not present in getIds), so I think it is good to make this visible to the user as is. 

More generally, _defaultPixelValue is derived from defaultLabelColor, defaultLabelVisibility, and defaultLabelOpacity. In addition to the above cases, defaultLabelColor is also applied in the following cases:
- The labelColor configuration is broken - this will be prevented by JSON schema validation in the future
- Missing table (when labelColor is configured as from/groupby) - this should be prevented by the user interface
- Unknown color map/palette (when labelColor is configured as from/groupby/random) - this should be prevented by the user interface
- Labels missing in the table (when labelColor is configured as from) or the colormap (derived from a table; iff the colormap's defaultLabelColor is not set)

I would argue that the last case should also be made visible to the user, but making this transparent could be defensible, too. What do you think?

Note that a Labels entry in the model defaults to `labelColor: { random: { palette: defaultLabelColorPalette } }`, so a missing `labelColor` configuration will always default to random colors.

Related: `constants.ts` has a TODO to replace `defaultLabelColor` with `defaultLabelColorPalette`. If we do that, "default" becomes a palette color and missing/failed labels become indistinguishable from resolved ones, so that decision should be made together with this one.

packages/@tissuumaps-core/src/model/constants.ts
jwindhager: Discussed offline:
- Single-channel images (sizeC === undefined) can yield grayscale or color tile sources
- Multi-channel images (sizeC > 0) require a transfer function value -> RGBA
- If all channel colors are undefined, apply a default color palette (global)
- If channel colors are only partially defined, show the undefined as grayscale (i.e., white)
- Use the QuPath color palette (uses RGB for first three colors; see https://github.com/qupath/qupath/blob/main/qupath-core/src/main/java/qupath/lib/images/servers/ImageChannel.java#L138-L160)

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonLabelsRenderer.ts
jwindhager: Discussed offline: don't hide. Hiding can work via visibilityMap/visibilityFrom (need to look into this later).

packages/@tissuumaps-core/src/model/constants.ts
jwindhager: Updates:

> Multi-channel images (sizeC > 0) require a transfer function value -> RGBA

Not true. Multi-channel image data objects that define getChannelData are expected to provide single samples (i.e., "grayscale values") per pixel that are then contrast-stretched & colorized for rendering; if getChannelData remains undefined, they are technically free to return whatever tile source (grayscale or RGB) they want.

> * If all channel colors are undefined, apply a default color palette (global)
> * If channel colors are only partially defined, show the undefined as grayscale (i.e., white)

We now always apply the default color palette. If data providers only provide partial colors, it is their job to return a default color for the remaining ones.

jwindhager: Documented this on the ImageData interface, resolving.

cavenel (approved):

PR #178, opened 2026-09-08
feat: let plugins contribute a user interface panel

jwindhager:

Jira: [TMAP-199](https://scilifelab.atlassian.net/browse/TMAP-199)

Fills in the `## User interface plugins` TODO in `apps/docs/docs/development/plugins.md`.

# Type of change

- [ ] Other

# List of changes made

Plugins could read and write all state, but had no way to show anything: the dockview layout was a static component map and a one-shot `onDockviewReady`.

- **Plugin interface** (`@tissuumaps/core`): `id`, `name`, `setup?` and `mount?`. `setup(stores)` may return a teardown callback; `mount(container, stores)` may return an unmount callback. The explicit `teardown` property is gone (breaking), and `setup` is optional — a plugin that only adds a user interface needs no `setup`, since `mount` receives the same stores. `mount` is a plain DOM contract, so plugins keep working as `<script>`s and core stays React-free.
- **Lifecycle**: `setup`, `mount`, `unmount`, `teardown`, all optional. If `setup` throws, the plugin is not registered and nothing is torn down; if `mount` throws, the teardown callback returned by `setup` is still called. Errors from plugin code are caught and logged everywhere.
- **The plugin registry owns the lifecycle** (`apps/tissuumaps/src/plugins.ts`). It creates the container element, calls `setup` and `mount`, keeps the unmount and teardown callbacks, and writes `appStore.plugins` directly via `setState` — the `registerPlugin`/`unregisterPlugin` store actions are removed. A plugin lands in the store, together with its `container`, only after `setup` and `mount` succeeded, and is removed before `unmount` and `teardown` run. Since the registry writes `HTMLElement`s into the Immer store, the entry is wrapped in `castDraft`; Immer leaves DOM nodes alone at runtime.
- **Panels are derived from `appStore.plugins`**. `usePluginPanels` reconciles `plugin:<id>` dockview panels against the registered plugins that have a `mount` (add / remove / rename), and a single `PluginPanel` bridge component attaches the plugin's container element. Mounting happens at registration, not when the panel is shown, so the user interface survives tab switches, drags and panel remounts.
- Closing the panel unregisters the plugin, via `DockviewDefaultTab`'s `closeActionOverride`. Intercepting the click rather than observing `onDidRemovePanel` matters: the latter also fires when dockview disposes its panels (StrictMode's dev double-mount, HMR) and would silently unregister every plugin.
- Panels are added with `renderer: "always"` so switching tabs does not remount them, and `inactive: true` so a plugin registering during startup does not steal the active tab.

# Screenshot of the fix

# Use of AI

Claude Code was used for the design, implementation, documentation and a review pass over the diff.

# Testing

Instructions on how to test — with `pnpm dev` running, paste into the browser console:

```js
window.tissuumaps.registerPlugin({
  id: "demo",
  name: "Demo",
  setup: () => {
    console.log("setup");
    return () => console.log("teardown");
  },
  mount: (container, { projectStore }) => {
    container.textContent = `${projectStore.getState().images.length} images`;
    return () => console.log("unmount");
  },
});
```

`setup` should log immediately, a "Demo" tab should appear in the sidebar group without stealing focus from Project, survive tab switching and dragging without remounting, and disappear — logging `unmount` and then `teardown`, in that order — both on `window.tissuumaps.unregisterPlugin("demo")` and when its close button is clicked.

# Further comments

`mount` is called with a container that is not yet part of the document; TissUUmaps attaches it to the panel once that is shown. Plugins that need their size should use a `ResizeObserver` rather than measuring once in `mount`. Re-registering a plugin under the same `id` unregisters the previous one first and mounts the new one into a fresh container.

### DIFF
--- a/apps/docs/docs/development/plugins.md
+++ b/apps/docs/docs/development/plugins.md
@@ -57,16 +57,20 @@ A plugin is unregistered again — tearing it down — using
 - `id` (required): a unique identifier for the plugin. Registering a plugin whose
   `id` is already in use unregisters the previous plugin first.
 - `name` (required): a human-readable name for the plugin.
-- `setup` (required): called once, immediately upon registration, with a reference
+- `setup` (optional): called once, immediately upon registration, with a reference
   to each of the application's Zustand stores (see below). Errors thrown by
   `setup` are caught and logged; they do not abort application startup. The
   plugin is then not registered, and its `teardown` is _not_ called - a
   `teardown` never has to cope with a half-initialized plugin. A `setup` that can
   fail part-way through is responsible for releasing what it had already set up
-  before it rethrows.
+  before it rethrows. A plugin that only adds a user interface does not need a
+  `setup`, because its panel's `mount` receives the same stores.
+- `panel` (optional): mounts a user interface panel shown for as long as the
+  plugin is registered, see [User interface plugins](#user-interface-plugins).
 - `teardown` (optional): called when the plugin is unregistered, and when the
-  application shuts down. Only ever called for a plugin whose `setup` returned
-  successfully. Errors thrown by `teardown` are caught and logged.
+  application shuts down. Only ever called for a plugin that was registered
+  successfully, i.e. one whose `setup` did not throw. Errors thrown by `teardown`
+  are caught and logged. It is called _before_ the plugin's `panel` is unmounted.
 
 ## Stores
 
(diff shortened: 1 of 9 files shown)


jwindhager:
Note to myself: we may want to give plugins more dockview-related options (e.g., should the tab be closable? where should the panel be placed?).

EDIT: Not for now, maybe in a later PR.

jwindhager:
Discussed offline: perhaps double-check/remove docs on width/height estimation

PR #179, opened 2026-09-09
Feat/tiff image data provider

cavenel:

- Builds on the multi-channel image rendering (#170), the OpenSeadragon data-transfer pipeline (#177) and the histogram-based default contrast limits (#191), all already in `development`.
- `geotiff-tilesource` is taken from the `tissuumaps` branch of TissUUmaps/GeoTIFFTileSource (a fork of pearcetm/GeoTIFFTileSource), pinned to `fcea2d0`; each change is also opened upstream on its own branch.
- `geotiff` 3.0.5 carries two upstream fixes as a pnpm patch: geotiffjs/geotiff.js#546 (LZW dictionary one code short) and geotiffjs/geotiff.js#536 (deferred tag arrays read little-endian regardless of byte order).

# Type of change

- [ ] Other

# List of changes made

**`@tissuumaps/storage` — TIFF image data provider** (data source type `tiff`, fields `url`/`path`, `z`, `t`)

- Reads OME-TIFF, QPTIFF (PerkinElmer/Akoya) and plain pyramidal TIFF (incl. SVS-style whole-slide files that are not JPEG 2000) over HTTP range requests or from a workspace file handle, through `openTIFF`. Opening a file reads only its directories and one small pyramid level per channel.
- `findTIFFParser` picks one of three `TIFFParser` implementations, which tell full-resolution planes from reduced copies: `OMETIFFParser` (OME-XML `TiffData` mapping with `z`/`t` selection and multi-file rejection), `QPTIFFParser` (per-IFD XML, thumbnail/overview/label dropped) and `PlainTIFFParser` (by size, accepts any file). Pyramids come from SubIFDs when present (geotiff.js has no SubIFD support, so directories are parsed through its parser), otherwise from same-width IFD groups; label and macro images are filtered by aspect ratio. A single RGB, YCbCr, palette or white-is-zero plane is drawn in its own colors; anything else with more than one sample per pixel is rejected with a clear error.
- Channel names and colors come from the file; when a file colors only some channels, the rest are left uncolored so that the renderer's index-based defaults never mix with metadata colors.
- Contrast limits are not estimated here: `readChannelHistograms` reads a 1024-bin value histogram per channel from a ~262144-pixel, tile-aligned, centred crop of the smallest sufficient level, and the renderer derives the limits from it (#191). Integers of 8 bits or fewer get no histogram and no pixel reads at all — the renderer stretches them over their data type range. At most one channel per decoder worker is read at a time.
- Tiles are served by `geotiff-tilesource` as raw rasters; `TIFFImageData.getTileData` hands a channel's band to the renderer's data transfer, which colorizes and contrast-stretches it. `copyRasters: false` hands out the cached raster itself (halves per-tile memory; nothing mutates it). A shared decoder pool (≤ 4 workers) runs this package's own worker so the patched decoders are used off the main thread too.

**Build**

- `geotiff` + `geotiff-tilesource` dependencies; `patches/geotiff@3.0.5.patch` via `patchedDependencies`; worker bundles disable code splitting so the inlined worker is self-contained; `geotiff`, `geotiff-tilesource`, `openseadragon` externalised; storage's `openseadragon` peer range aligned with core/render (`^6.0.2`).

**App**: `TIFFImageDataProvider` registered as a built-in provider.

**Docs**: `supported-data/tiff.md`; `development/dependencies.md` gains "Patched dependencies" and "Forked dependencies" sections; `development/code-architecture.md` records that format metadata belongs in the data provider rather than in the tile source.

# Screenshot of the fix

None attached.

# Use of AI

Claude Code was used for implementing the TIFF tile source. The code was then reviewed.

# Testing

- A rebuild is needed due to new dependencies: `pnpm install` (new `geotiff`, `geotiff-tilesource` GitHub tarball, patched `geotiff`; `pnpm-lock.yaml` pins everything, no extra steps).
- Unit tests: `pnpm --filter @tissuumaps/storage test` — 50 tests under `src/tiff` (structure parsers incl. SubIFDs, OME `TiffData`/`DimensionOrder`/UUID handling, color defaults, histogram windows, bins and concurrency, plus two on real files written with geotiff.js).
- Manual, in the app (Images → add → type **TIFF**):
  1. OME-TIFF with SubIFD pyramid via URL: channels named and coloured from the file, contrast sensible without any project config; `z`/`t` select planes.
  2. QPTIFF: channel names/colours from the QPI descriptions; thumbnail/overview/label not shown as levels.
  3. Aperio SVS or other plain pyramidal RGB TIFF: opens as a single RGB image.
  4. Plain single-channel grayscale TIFF: renders white, not in an index-based colour.
  5. Non-pyramidal strip-stored TIFF: opens (slow to pan, as documented); the open-time histogram read is quick.
  6. Same files via a workspace `path`.
  7. `z: 1` on a QPTIFF or plain TIFF → "z and t require an OME-TIFF file." error.
- Remote files need a server that supports HTTP range requests.

# Further comments

- Layout detection lives in this repo, not in the plugin: the plugin's own `getAllTileSources`/`resolveLayout`/QPTIFF parsing is unused by TissUUmaps (it has no OME-TIFF support and no per-channel metadata API). Whether to drop those paths from the `tissuumaps` fork branch is a fork decision.
- The geotiff patch is applied by this workspace's `pnpm install`, not by installing `@tissuumaps/storage`, and `geotiff` is external — so a downstream project gets unpatched geotiff everywhere outside the inlined worker (big-endian deferred tags, LZW past 4093 codes). Documented in `dependencies.md` as a known limitation; it resolves itself when geotiffjs/geotiff.js#536 and #546 are released. The app in this repository is unaffected.
- Files with thousands of IFDs (OME z-stacks) still pay a serial directory walk at open; `perf/tiff-directory-reads` addresses it.
- `copyRasters: false` relies on no consumer mutating the values returned by `getTileData`; both `TIFFImageData` and the core contract say so.
- The `openseadragon` peer range of `@tissuumaps/storage` narrows from `>=5.0.1 <7.0.0` to `^6.0.2`; core and render already required that.


packages/@tissuumaps-storage/src/tiff/TIFFImageDataProvider.ts
```diff
@@ -0,0 +1,210 @@
...
+        type: "integer",
+        minimum: 0,
+      },
+      t: {
+        type: "integer",
+        minimum: 0,
+      },
+    },
+    required: ["url"], // TODO ... or path
+  };
```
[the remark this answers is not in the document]
cavenel: Will fix everywhere in the repo at once, no fix in this PR. Marked as TODO

packages/@tissuumaps-storage/src/tiff/formats/OMETIFFParser.ts
```diff
@@ -0,0 +1,238 @@
...
+  if (!dimensionOrders.includes(dimensionOrder)) {
+    throw new Error(`Invalid DimensionOrder "${dimensionOrder}".`);
+  }
+  const tiffData = XMLUtils.getChildren(pixels, "TiffData").map((td) => {
+    const uuid = XMLUtils.getChildren(td, "UUID")[0];
+    if (uuid !== undefined && uuid.textContent?.trim() !== fileUUID) {
+      const fileName =
+        uuid.getAttribute("FileName") ?? uuid.textContent?.trim();
+      throw new Error(`Multi-file OME-TIFF is not supported (${fileName}).`);
+    }
```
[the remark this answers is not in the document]
cavenel: Resolved differently, by changing the error message if `fileUUID === null`

apps/docs/docs/development/code-architecture.md
```diff
@@ -125,6 +125,8 @@ Each data provider has its own dedicated directory and is separately exported in
 
 Heavy parsing and decoding runs off the main thread: Parquet and GeoJSON in dedicated web workers (`parquet.worker.ts`, `geojson.worker.ts`, inlined into the bundle), CSV via PapaParse's worker mode.
 
+Format metadata belongs in the data provider, not in the tile source. The provider reads the file's channels, colors and pyramid levels itself and hands the tile source explicit images to decode, so that channel count, names, colors and contrast limits come from the file and can be unit-tested here. The TIFF provider works this way.
```
jwindhager: Not sure this is accurate. Format metadata should indeed be extracted by the data provider, but it may re-use the (custom) tile source to do so to avoid code duplication/divergence. For example, the OMEZarrImageDataProvider reaches into the OMEZarrTileSource to use the same chunk for computing the histogram as gets displayed in the viewer.

apps/docs/docs/supported-data/tiff.md
```diff
@@ -3,3 +3,86 @@ sidebar_position: 2
 ---
 
 # TIFF
+
+The built-in **TIFF data provider** opens OME-TIFF, QPTIFF and plain pyramidal TIFF files as **images**. Tiles are decoded in the browser with [geotiff.js](https://geotiffjs.github.io/), so no server-side tiling is needed; remote files need a server that supports HTTP range requests.
```
jwindhager: We also support non-pyramidal.

Also, should we perhaps have a way to override the auto-detection/force a variant (e.g. a `variant?: "plain" | "ome-tiff" |  "qptiff"` field in TiffImageDataSource)?

apps/docs/docs/supported-data/tiff.md
```diff
@@ -3,3 +3,86 @@ sidebar_position: 2
...
+
+Per-channel rendering settings come from the file where it records them:
+
+| Setting         | Source                                    | Fallback                                              |
+| --------------- | ----------------------------------------- | ----------------------------------------------------- |
+| Name            | OME-XML `Channel` `Name`, QPTIFF `Name`   | none                                                  |
+| Color           | OME-XML `Channel` `Color`, QPTIFF `Color` | a color chosen by the viewer                          |
+| Contrast limits | `[0, 255]` for 8-bit channels             | quantiles of a histogram read from a sample of pixels |
+| Visibility      | not recorded by TIFF                      | visible                                               |
```
jwindhager: The table still doesn't align with the OME-Zarr one :) I think this should only list fallbacks that are specific to the provider. In the OME-Zarr docs we have a paragraph after the table that says that the viewer defaults apply.

apps/docs/docs/supported-data/tiff.md
```diff
@@ -3,3 +3,86 @@ sidebar_position: 2
...
+    }
+  ]
+}
+```
+
+Channel settings that are left out fall back to the file's metadata and to the estimated contrast limits.
+
+## Limitations
+
+- Only one plane (`z`, `t`) of an OME-TIFF is shown; there is no in-app plane selection yet.
```
jwindhager: Well, one can configure the tile source...

packages/@tissuumaps-storage/src/tiff/formats/OMETIFFParser.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    const ifdByPlane = mapPlanesToIFDs(pixels, images.length);
+    const planes = Array.from({ length: getPlaneCountC(pixels) }, (_, c) => {
+      const ifd = ifdByPlane.get(planeKey(c, z, t));
+      const image = ifd !== undefined ? images[ifd] : undefined;
+      if (image === undefined) {
+        throw new Error(`No IFD for plane c=${c}, z=${z}, t=${t}.`);
+      }
+      return image;
```
jwindhager:
```suggestion
      if (ifd === undefined) {
        throw new Error(`No IFD for plane c=${c}, z=${z}, t=${t}.`);
      }
      return images[ifd];
```

packages/@tissuumaps-storage/src/tiff/formats/PlainTIFFParser.ts
```diff
@@ -0,0 +1,39 @@
...
+    const { full, reduced } = TIFFUtils.partitionBySize(images);
+    const pyramids =
+      (await TIFFUtils.readSubIFDPyramids(tiff, full, { signal })) ??
+      TIFFUtils.groupByWidth(full, reduced);
+    TIFFUtils.validatePyramids(pyramids);
+    return {
+      pyramids,
+      channels: TIFFUtils.hasOwnColors(pyramids)
+        ? undefined
+        : full.map(() => ({})),
```
jwindhager, line 38: What does this do?

packages/@tissuumaps-storage/src/tiff/formats/OMETIFFParser.ts
```diff
@@ -0,0 +1,253 @@
...
+    tiff: GeoTIFF,
+    options?: { signal?: AbortSignal },
+  ): Promise<boolean> {
+    const { signal } = options ?? {};
+    signal?.throwIfAborted();
+    const description = await TIFFUtils.readFirstDescription(tiff, {
+      signal,
+    });
+    const root = XMLUtils.parse(description);
+    return root?.localName === "OME";
```
jwindhager: Can we make this a constant, like for QPTIFF?

packages/@tissuumaps-storage/src/tiff/formats/TIFFParser.ts
```diff
@@ -0,0 +1,87 @@
...
+   * are drawn in their own colors
+   */
+  pyramids: GeoTIFFImage[][];
+
+  /** `undefined` for files that are drawn in their own colors */
+  channels: TIFFChannelMetadata[] | undefined;
+};
+
+/** Reads the structure of one TIFF flavor */
+export type TIFFParser = {
```
jwindhager: This should be an interface, I think?

packages/@tissuumaps-storage/src/tiff/formats/TIFFParser.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+/**
+ * The parsers in detection order
+ *
+ * The plain parser accepts any file, so it has to stay last: ahead of another
+ * parser it would claim that parser's files and drop their channel names and
+ * colors, without failing.
+ */
+const tiffParsers: TIFFParser[] = [
```
jwindhager: Maybe nice to export this?

packages/@tissuumaps-storage/src/tiff/geotiff-tilesource.d.ts
```diff
@@ -0,0 +1,40 @@
+// geotiff-tilesource ships no types; only the parts used here are declared
+declare module "geotiff-tilesource" {
```
jwindhager, line 3: Perhaps leave a comment that these were auto-generated using Claude Code for geotiff-tilesource 3.0.5?

packages/@tissuumaps-storage/vite.config.ts
```diff
@@ -17,6 +17,12 @@ export default defineConfig(({ mode }) => ({
   ],
   worker: {
     format: "es",
+    // inlined workers cannot load chunks, so they get no code splitting
```
jwindhager: Could you elaborate here? This is a geotiff.js-specific thing, right? 

This is what Claude Code had to say (as reference - don't make it that long please 😅):

> That makes the comment much more concrete. geotiff.js is close to a worst case for this, because it relies on dynamic imports in exactly the code that ends up inside the worker.
> 
> Two things collide:
> 
> 1. geotiff's Pool runs decoding in workers. Pool spins up src/worker/decoder.js to decode tiles off the main thread. If you're bundling that worker inline (blob/data URL), you get the no-base-URL problem from before.
> 
> 2. geotiff loads decoders lazily. Its compression registry maps TIFF Compression tag values to () => import('./lzw.js'), () => import('./jpeg.js'), () => import('./deflate.js'), () => import('./lerc.js'), and so on. That's deliberate — most people only ever hit one or two codecs, and LERC in particular drags in a big wasm payload nobody wants unconditionally.
> 
> So the worker's module graph is full of dynamic imports. With code splitting on, Rolldown emits each decoder as its own chunk, and the inlined worker tries to fetch ./lzw-abc123.js relative to blob:https://yoursite/…. Nothing resolves. The symptom is nasty because it's data-dependent: uncompressed TIFFs decode fine, then someone loads a DEFLATE-compressed COG and you get a failed dynamic import inside the worker — often surfacing as a hung or rejected pool.decode() rather than an obvious module error, since it happens across the worker boundary.
> 
> codeSplitting: false collapses all of that into the single worker file, so every decoder is present from the start.
> 
> The tradeoff is the one you'd expect: your inlined worker now carries every codec geotiff supports, LERC's wasm included. That can be a few hundred KB you didn't ask for. If it matters, the options are to stop inlining the worker (let it be a real emitted file with sibling chunks), or skip Pool entirely and decode on the main thread, which is fine for small rasters and avoids the whole question.

packages/@tissuumaps-storage/src/tiff/XMLUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+  static getChildren(element: Element, tagName: string): Element[] {
+    return [...element.children].filter((child) => child.localName === tagName);
+  }
+
+  /**
+   * Returns the trimmed text of the first direct child with the given name
+   *
+   * @param element - The parent
+   * @param tagName - The child's local name
+   * @returns The text, or `undefined` if there is no such child or it is empty
```
jwindhager:
```suggestion
   * @returns The trimmed text, or `undefined` if there is no such child or it is empty
```

packages/@tissuumaps-storage/src/tiff/XMLUtils.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Returns an integer attribute
+   *
+   * @param element - The element
+   * @param name - The attribute name
+   * @param defaultValue - Returned if the attribute is missing or not an integer
+   * @returns The value
+   */
+  static getIntAttribute(
```
jwindhager, line 56: Do we really need this? Isn't `NumberUtils.tryParseFinite(element.getAttribute(name)) ?? defaultValue` sufficient?

packages/@tissuumaps-storage/src/tiff/TIFFUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+    if (offsetsPerPlane.some((offsets) => offsets.length !== count)) {
+      throw new Error(
+        `The channels have different numbers of SubIFDs (${offsetsPerPlane.map((o) => o.length).join(", ")}).`,
+      );
+    }
+    if (count === 0) {
+      return undefined;
+    }
+    // geotiff.js has no SubIFD support, so the directories are parsed the way
+    // GeoTIFF.getImage() does it
```
jwindhager, line 111: What does `GeoTIFF.getImage()` refer to? If this has SubIFD support, why don't we use it here?

packages/@tissuumaps-storage/src/tiff/TIFFUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+
+  /**
+   * Checks that the pyramids can be displayed
+   *
+   * @param pyramids - One pyramid per plane, largest image first
+   * @throws Error if there is no pyramid, if the levels of a pyramid do not
+   * shrink, or if the image does not draw in its own colors and a plane is not
+   * a single black-is-zero band
+   */
+  static validatePyramids(pyramids: GeoTIFFImage[][]): void {
```
jwindhager, line 242: Why choose a plural version over a singular `validatePyramid`?

packages/@tissuumaps-storage/src/tiff/TIFFUtils.ts
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  /**
+   * Reads the `ImageDescription` of every image
+   *
+   * @param images - The images
+   * @param options - An abort signal
+   * @returns One description per image
+   */
+  static readDescriptions(
```
jwindhager: Do we need the plural version, or would the singular be enough?

jwindhager (changes requested):
Really nice work, thanks!

Only a few minor changes/clarifications.

packages/@tissuumaps-storage/src/tiff/TIFFUtils.ts
cavenel: It takes all the pyramids and two of its checks are cross-pyramid (empty set, and the own-colours case, which looks at how many pyramids there are). A singular version couldn't check those.

packages/@tissuumaps-storage/src/tiff/TIFFUtils.ts
cavenel: Inlined into QPTIFFParser and deleted; it was only a Promise.all over readDescription, which throws on abort by itself.

packages/@tissuumaps-storage/src/tiff/TIFFUtils.ts
cavenel: GeoTIFF.getImage() is geotiff.js's own method, which opens an image by its index in the file's top-level IFD chain and wraps it in a GeoTIFFImage. It **cannot** open a SubIFD, because SubIFDs are not in that chain: the tag gives byte offsets, and no geotiff.js method builds an image from an offset. So we do it by hand: read the offsets from the SubIFDs tag, parse the directory at each one, and build the GeoTIFFImage the same way getImage() does.

packages/@tissuumaps-storage/src/tiff/XMLUtils.ts
cavenel: `tryParseFinite` would allow floats, we don't want that, but I replaced with `tryParseSafeInt`.

packages/@tissuumaps-storage/src/tiff/XMLUtils.ts
cavenel: Done.

packages/@tissuumaps-storage/vite.config.ts
cavenel: Reworded the comment: "geotiff.js's worker imports its decoders dynamically; an inlined worker cannot fetch chunks, so workers are built as a single file."
That's the concrete case: each codec is a () => import(...), and inside a blob-URL worker those chunks don't resolve, so it fails only on compressed files. The setting is package-wide, so I didn't word it as geotiff-only.

packages/@tissuumaps-storage/src/tiff/geotiff-tilesource.d.ts
cavenel: Added.

packages/@tissuumaps-storage/src/tiff/formats/TIFFParser.ts
cavenel: Exported now as a module-level export. Or did you mean also exported in `index.ts`?

packages/@tissuumaps-storage/src/tiff/formats/TIFFParser.ts
cavenel: Yes. Changed.

packages/@tissuumaps-storage/src/tiff/formats/OMETIFFParser.ts
cavenel: Done.

packages/@tissuumaps-storage/src/tiff/formats/PlainTIFFParser.ts
cavenel: `channels` says whether the file is channel data, and how much is known about each channel.

`hasOwnColors` is true when there is one pyramid and its full-resolution plane already carries colors: RGB or YCbCr with 3 or 4 samples, a palette image, or white-is-zero grayscale. The tile source draws those as they are, so there are no channels and channels is undefined.

Otherwise every pyramid is a channel, and there is one entry per channel. Plain TIFF records no channel name or color anywhere, so the entries are empty, the count is the only information the file gives. OME-TIFF and QPTIFF fill the same entries from their XML.

packages/@tissuumaps-storage/src/tiff/formats/OMETIFFParser.ts
cavenel: Applied, with one change: `images[ifd]` needs a `!`. We have `noUncheckedIndexedAccess` on, so TypeScript treats any indexed access as possibly undefined, and this has to return a `GeoTIFFImage`. The index is always valid, because `mapPlanesToIFDs` only maps IFDs the file has.

apps/docs/docs/supported-data/tiff.md
cavenel: Removed the whole line from limitation (I also removed the line from the ome-zarr limitations 😄)

apps/docs/docs/supported-data/tiff.md
cavenel: Fixed.

apps/docs/docs/supported-data/tiff.md
cavenel: Fixed for non-pyramidal.
I would keep the "forcing variant" feature out of this PR so that we get the PR out. Could be a nice extra feature to add later.

apps/docs/docs/development/code-architecture.md
cavenel: Reworded: "Format metadata belongs in the data provider: channel count, names, colors and contrast limits come from the file, not from the viewer. How the provider reads it is up to it. OMEZarrImageDataProvider goes through OMEZarrTileSource, TIFFImageDataProvider parses the IFDs itself."

packages/@tissuumaps-storage/src/tiff/TIFFUtils.ts
jwindhager: Ok - I think the comment is a bit misleading. I read it as: geotiff.js doesn't support SubIFD, so we do it like geotiff.js (which doesn't make much sense). But all good :)

jwindhager (approved):

PR #184, opened 2026-09-11
Feat/d3 color palettes

cavenel:

Brings over the d3 color palettes offered by TissUUmaps 3 (`_d3LUTs`).

# Type of change

- [ ] Other

# List of changes made

- **`@tissuumaps/core`: 39 continuous and 12 categorical color palettes**, up from one each. Adds `d3-scale-chromatic` and `d3-color` as dependencies, externalized in the lib build like `gl-matrix`. Sampling the schemes at load costs ~6.5 ms; shipping them as `.txt` fixtures instead would have meant ~260 kB of generated assets and a generator script to maintain.
- **`ColorUtils.sampleColorPalette(interpolate, size)`** samples a continuous scheme into colors. The text-based palettes are untouched — Batlow still comes from its `.txt`, and `parseColorPalette` still reads it.
- **Continuous palettes are the 38 schemes from TissUUmaps 3**, at 256 colors each, ordered perceptually-uniform first then alphabetically. **Categorical palettes are d3's discrete schemes** (Category 10, Observable 10, Tableau 10, Accent, Dark 2, Paired, Pastel 1/2, Set 1/2/3) rather than sampled continuous ramps: groups are picked by hash, so neighbouring indices in a 256-step ramp would be indistinguishable.
- **Palette IDs are short and stable** (`viridis`, `category10`), since they are persisted in project configs. `batlow` and `batlowS` are unchanged, and the resolvers still look up the merged `colorPalettes`, so existing projects keep rendering as before.
- **The palette dropdown is split by data type**: continuous palettes when coloring from a column, categorical ones when coloring by group or at random. Each option previews the palette it names.
- **Fix: removed three inline styles in `SimpleSelect` that overrode Tailwind classes on the same elements.** The list's `maxHeight` was merged after Base UI's own, which sets `maxHeight: 100%` only while `alignItemWithTrigger` is active — so a long list stopped halfway down a full-height popup and scrolled erratically. The popup's `minWidth` likewise overrode its `data-[side=none]` width. Matches the upstream Base UI example, which sets no inline styles.

# Screenshot of the fix

# Use of AI

Claude Code was used to weigh the d3-dependency approach against generating `.txt` fixtures, and to write the palette tables. All of it was reviewed and adjusted before committing.

# Testing

- A rebuild is needed due to new dependencies (`pnpm install`)
- Unit tests cover `sampleColorPalette` and assert that every palette has a unique ID, at least two colors, and finite components
- To test by hand: color points by a numeric column and check the continuous list; color by a categorical column and check the categorical list. Open the dropdown and confirm it sizes to its content rather than the viewport.

# Further comments

A `groupBy` or `random` config that still points at a continuous palette ID renders correctly — the resolver looks up the merged list — but its select shows empty, because that ID is no longer in the categorical list it offers.

The `data-[side=none]` Tailwind variants in `simple-select.tsx` are kept as-is; the file is a near-verbatim adaptation of the Base UI example and the comment there records the deviations.


apps/tissuumaps/src/components/widgets/config/ColorConfigWidget/index.tsx
```diff
@@ -250,13 +252,10 @@ function GroupByColorConfigWidget({
       </Field>
       <Field disabled={map !== null}>
         <FieldLabel>Color palette</FieldLabel>
-        <SimpleSelect
-          items={colorPalettes}
-          itemLabel={(colorPalette) => colorPalette.name}
-          itemValue={(colorPalette) => colorPalette.id}
+        <ColorPaletteSelect
+          colorPalettes={categoricalColorPalettes}
```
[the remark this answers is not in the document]
cavenel: At this point we don't care about breaking old project, especially if not using correct palettes...

packages/@tissuumaps-core/src/utils/ColorUtils.ts
```diff
@@ -36,6 +38,30 @@ export class ColorUtils {
...
+   * @param interpolate - Maps a position within `[0, 1]` to a CSS color string
+   * @param size - Number of colors to sample, at least two
+   * @returns The sampled colors, in the order of the scheme
+   * @throws Error if `size` is smaller than two
+   */
+  static sampleColorPalette(
+    interpolate: (t: number) => string,
+    size: number,
+  ): Color[] {
+    if (size < 2) {
```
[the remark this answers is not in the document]
cavenel: Folded Number.isInteger into the existing guard

apps/tissuumaps/src/components/widgets/config/ColorConfigWidget/ColorPaletteSelect.tsx
```diff
@@ -0,0 +1,65 @@
...
+      itemLabel={colorPaletteLabel}
+      itemValue={colorPaletteId}
+      value={value}
+      onValueChange={onValueChange}
+      nullable
+    />
+  );
+}
+
+function colorPaletteLabel(colorPalette: ColorPalette) {
```
jwindhager: Names of components (functional or not) should be capitalized

apps/tissuumaps/src/components/widgets/config/ColorConfigWidget/ColorPaletteSelect.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <>
+      <ColorPaletteSwatch colors={colorPalette.colors} />
+      {colorPalette.name}
+    </>
+  );
+}
+
+function colorPaletteId(colorPalette: ColorPalette) {
```
jwindhager: I think this can be inlined

packages/@tissuumaps-core/src/palettes.ts
```diff
@@ -34,22 +86,99 @@ export type ColorPalette = {
   colors: Color[];
 };
 
+/** Number of colors sampled from each continuous D3 color scheme */
+const continuousColorPaletteSize = 256;
```
jwindhager: Perhaps better to call it `sampledColorPaletteLength`?

packages/@tissuumaps-core/src/palettes.ts
```diff
@@ -34,22 +86,99 @@ export type ColorPalette = {
...
+
+/** Builds a color palette by sampling a continuous D3 color scheme */
```
jwindhager: Please complete the function's TSDoc

packages/@tissuumaps-core/src/palettes.ts
```diff
@@ -34,22 +86,99 @@ export type ColorPalette = {
...
+
+/** Builds a color palette from a list of hex colors */
+function hexColorPalette(
+  id: string,
+  name: string,
+  hexColors: readonly string[],
+): ColorPalette {
+  return { id, name, colors: hexColors.map((hex) => ColorUtils.fromHex(hex)) };
+}
+
```
jwindhager, line 127: Both `sampledColorPalette` and `hexColorPalette` are trivial - could we inline those?

packages/@tissuumaps-core/src/utils/ColorUtils.ts
```diff
@@ -37,6 +39,35 @@ export class ColorUtils {
...
+   * whole `[0, 1]` range, both ends included.
+   *
+   * @param interpolate - Maps a position within `[0, 1]` to a CSS color string
+   * @param size - Number of colors to sample, an integer of at least two
+   * @returns The sampled colors, in the order of the scheme
+   * @throws Error if `size` is not an integer of at least two, or if the
+   * scheme yields a color string that cannot be parsed
+   */
+  static sampleColorScheme(
+    interpolate: (t: number) => string,
```
jwindhager:
```suggestion
    colorScheme: (t: number) => string,
```

packages/@tissuumaps-core/src/utils/ColorUtils.ts
```diff
@@ -37,6 +39,35 @@ export class ColorUtils {
...
+    size: number,
```
jwindhager:
```suggestion
    n: number,
```

packages/@tissuumaps-core/src/utils/ColorUtils.ts
```diff
@@ -37,6 +39,35 @@ export class ColorUtils {
...
+  ): Color[] {
+    if (!Number.isInteger(size) || size < 2) {
+      throw new Error(`Invalid color scheme size: ${size}`);
+    }
+    return Array.from({ length: size }, (_, i) => {
+      const spec = interpolate(i / (size - 1));
+      const { r, g, b } = parseCSSColor(spec);
+      if (Number.isNaN(r) || Number.isNaN(g) || Number.isNaN(b)) {
+        throw new Error(`Invalid color scheme color: ${spec}`);
+      }
```
jwindhager, line 69: Is this necessary? Doesn't d3-color's `rgb` guarantee valid numbers?

packages/@tissuumaps-render/src/resolvers/ColorResolver.ts
```diff
@@ -426,17 +431,28 @@ export class ColorResolver extends ResolverBase {
...
+      return color;
     }
-    return undefined;
+    const t = position - index;
+    return {
+      r: color.r + t * (nextColor.r - color.r),
+      g: color.g + t * (nextColor.g - color.g),
+      b: color.b + t * (nextColor.b - color.b),
+    };
   }
```
jwindhager, line 456: In principle, I think the interpolation is good, but the section is a (relatively verbose) rewrite of the original code (e.g. early abort for v, destructuring of colors, split of position/index). For next time, to make reviews easier, it would be nice to stick with the original code as close as possible. But good for now :)

jwindhager (changes requested):
Thanks! Only minor changes from my side :)

packages/@tissuumaps-core/src/palettes.ts
cavenel: They're called more than 50 times, do you mean repeating the same lines 50 times? I think I like the functions better!

packages/@tissuumaps-core/src/utils/ColorUtils.ts
cavenel: Does it cost much? I don't think d3 guarantees valid numbers no, so I would keep that in case!

packages/@tissuumaps-core/src/palettes.ts
cavenel: `sampledColorPaletteColorCount`?

packages/@tissuumaps-core/src/utils/ColorUtils.ts
jwindhager: No, but it would be incomplete (no check of 0<=x<=255), so I'm wondering what is checked here.

packages/@tissuumaps-core/src/palettes.ts
jwindhager: `sampledColorPaletteSize`? :)

packages/@tissuumaps-core/src/utils/ColorUtils.ts
cavenel: As d3-color reports a parse failure as NaN components, I think this should cover most cases!

jwindhager (approved):
<3

PR #185, opened 2026-09-11
fix(app): window the items data table

cavenel:
# Description

The annotations table built one row object per item and handed all of them to
react-table, so opening it for a large dataset cost time and memory proportional
to the item count. It now materializes only the rows within the visible range.

Windowing alone does not get past a second limit: browsers cap the height of an
element at a few ten million pixels and silently clamp anything taller, so with
36px rows a list stops scrolling after about 600,000 rows. Taller content is
therefore laid out compressed, and scroll positions are scaled between the
layout and the content.

Compression breaks TanStack Virtual's assumption that a scroll position is a
content position, which the first version worked around by replacing the
virtualizer's scroll observation and correcting every row position after it.
With a fixed row height the rest of the virtualizer is arithmetic, so the table
now computes the visible range itself and the dependency is gone.

None.

# Type of change

- [ ] Other

# List of changes made

- Only the rows within the visible range are built, from the item ids and
  per-index accessors, so the cost of the table no longer depends on the item
  count.
- Rows have a fixed height of 36px and are clipped to it. Cells of extra columns
  must fit within it.
- Content taller than 10,000,000px is laid out at that height, and every scroll
  position is scaled by the ratio between the two.
- A new `VirtualTable` in `components/common/` owns the scrolling, the
  compression and the rendering. It takes a row count and a window of rows, and
  knows nothing about annotations.
- The annotations table is split into an item table and a group table, each with
  its own hook loading its rows, instead of one component switching on a union
  of both row types.
- The groups are kept together with the table and column they were loaded from,
  rather than cleared from the effect that loads them.
- `@tanstack/react-table` upgraded from v8 to v9: features are registered
  explicitly and the core row model option is gone. This also removes the
  `"use no memo"` directive, as v9 fixes the React Compiler incompatibility
  (TanStack/table#5567). Three calls were dropped as dead: cell placeholders and
  cell visibility belong to features the table does not register, and no header
  renders a resize handle.
- `@tanstack/react-virtual` removed.

# Screenshot of the fix

Not relevant.

# Use of AI

Claude Code was used.

# Testing

- Open a project with a large points dataset and open its Data panel. Scrolling
  should be smooth and memory should not grow with the item count.
- Scroll a table with more than ~600,000 rows to the very bottom. The last row
  must be reachable, and the rows must stay aligned with the scrollbar
  throughout. This is the path the compression covers and the one worth the most
  attention.
- Scroll the table, switch to another tab and back. The table should show the
  same rows as before.
- Select a table and a group-by column. The group rows should list and scroll as
  before, with the per-group marker, size, color, visibility and opacity
  columns.
- Resize the panel while scrolled into the middle of a large table.

# Further comments

The branch was rebased onto development, so the Copilot review above points at
commits that no longer exist. Its finding, that rows must be clipped to their
fixed height, is addressed by `fix(app): clip items data table rows to their
fixed height`.

The compression means the scrollbar cannot address individual rows on very large
tables: at 100M rows one layout pixel covers several hundred rows. Going to a
specific row will need an index input or a filter rather than a scroll, and
`VirtualTable` will need to hold the content offset as state instead of deriving
it from `scrollTop`. Out of scope here.


cavenel:
Apparently, a Chrome div has a max height of 22.4 million px, which means a max of 621,377 rows × 36 px. We can have millions of rows when not grouped. So this PR doesn't completely fix that...

apps/tissuumaps/src/components/widgets/AnnotationsWidget/useItemRows.ts
```diff
@@ -0,0 +1,74 @@
...
+    let annotatedIds: Set<number> | undefined;
+    if (data !== undefined) {
+      ids = data.getIds();
+      if (table !== null) {
+        // the selected table governs the names; while it is still loading there
+        // are none yet, rather than the object's own names, which would show a
+        // different column for a moment and then be replaced
+        if (tableData !== null) {
+          const tableIds = tableData.getIds();
+          annotatedIds = new Set(tableIds);
```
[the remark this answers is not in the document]
cavenel: `tableIds === ids` now sets an `allAnnotated` flag instead of building the set; the set is only constructed for item types whose ids differ from the table's. The reference check was already being used a few lines below to align names by index, so this just applies the same fact consistently.

apps/tissuumaps/src/components/widgets/AnnotationsWidget/useGroupRows.ts
```diff
@@ -0,0 +1,62 @@
...
+    };
+  }, [tableData, groupByColumn]);
+
+  return useMemo(() => {
+    if (groups === null) {
+      return null;
+    }
+    const groupRows = groups.map((group) => ({ group: String(group) }));
+    groupRows.sort((a, b) => a.group.localeCompare(b.group));
+    return groupRows;
```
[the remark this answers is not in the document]
cavenel: Fixed. The hook now keeps the loaded values as plain strings, sorts those, and wraps only the requested range into row objects inside getRows. It returns { rowCount, getRows, loaded }, matching useItemRows, so both modes window the same way.

apps/tissuumaps/src/components/common/virtual-table.tsx
```diff
@@ -0,0 +1,341 @@
...
+  contentSize: number,
+  viewportSize: number,
+  maxContentSize: number,
+): { layoutSize: number; factor: number } {
+  const range = contentSize - viewportSize;
+  const maxRange = maxContentSize - viewportSize;
+  if (range <= maxRange || maxRange <= 0) {
+    return { layoutSize: contentSize, factor: 1 };
+  }
+  return { layoutSize: maxContentSize, factor: range / maxRange };
```
jwindhager: Can we rename `factor` to something more descriptive here? E.g. `layoutToContentFactor`?

apps/tissuumaps/src/components/common/virtual-table.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+/**
+ * The tallest scroll content that is laid out, in pixels
+ *
+ * Browsers cap the height of an element at a few ten million pixels, Chrome
+ * at about 33 million divided by the page zoom, and clamp anything taller, so
+ * that a longer list could not be scrolled past the cap. Content beyond this
+ * height is compressed instead, see {@link compressScrollRange}.
+ */
+const maxScrollContentHeight = 10_000_000;
```
jwindhager: Can we make this a prop with default value?

apps/tissuumaps/src/components/common/virtual-table.tsx
```diff
@@ -0,0 +1,341 @@
...
+
+/**
+ * How many rows are rendered beyond each end of the visible range
+ *
+ * Rows outside of it are not rendered, so scrolling reveals blank space until
+ * the next render; the overscan covers a scroll of up to this many rows.
+ */
+const overscan = 2;
```
jwindhager: Can we make this a prop with default value?

apps/tissuumaps/src/components/common/virtual-table.tsx
```diff
@@ -0,0 +1,341 @@
...
+  /** Attached to the scroll container */
+  containerRef: RefObject<HTMLDivElement | null>;
+  /** Attached to the header that sticks to the top of the scroll container */
+  headerRef: RefObject<HTMLTableSectionElement | null>;
+  /** The first row within the visible range */
+  firstIndex: number;
+  /** The row past the last one within the visible range */
+  lastIndex: number;
+  /** The height to lay the rows out at, in pixels */
+  rowsHeight: number;
```
jwindhager: Should we name this `contentHeight`, to stick with the terminology in `compressScrollRange`? Or is this something different?

apps/tissuumaps/src/components/common/virtual-table.tsx
```diff
@@ -0,0 +1,341 @@
...
+ *
+ * @param rowCount - The number of rows in the list
+ * @param rowHeight - The height of every row, in pixels
+ * @param height - The height of the scroll container, in pixels
+ * @returns The refs to attach, the visible range of rows and their layout
+ */
+function useCompressedVirtualizer(
+  rowCount: number,
+  rowHeight: number,
+  height: number,
```
jwindhager: Same as `contentHeight`?

apps/tissuumaps/src/components/common/virtual-table.tsx
```diff
@@ -0,0 +1,341 @@
...
+): CompressedVirtualizer {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const headerRef = useRef<HTMLTableSectionElement>(null);
+  const [headerHeight, setHeaderHeight] = useState(0);
+  const [viewportHeight, setViewportHeight] = useState(height);
```
jwindhager: Same as `contentHeight`?

apps/tissuumaps/src/components/common/virtual-table.tsx
```diff
@@ -0,0 +1,341 @@
...
+ * offset alone. Past the height a browser lays out, the rows are compressed
+ * into the height it does, and every row is shifted by how far the list has
+ * run ahead of the layout at the current scroll offset.
+ *
+ * @param rowCount - The number of rows in the list
+ * @param rowHeight - The height of every row, in pixels
+ * @param height - The height of the scroll container, in pixels
+ * @returns The refs to attach, the visible range of rows and their layout
+ */
+function useCompressedVirtualizer(
```
jwindhager: This is a potentially reusable hook and should go to the hooks folder, right? Also, if it is specific to *rows* stacked vertically, I would call it `useCompressedRowVirtualizer`.

apps/tissuumaps/src/components/widgets/AnnotationsWidget/AnnotationsGroupTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+        });
+    }
+    return () => {
+      abortController.abort();
+    };
+  }, [tableData, groupByColumn]);
+
+  // the groups are sorted as the plain values they are loaded as; a row object
+  // per group would cost as much as one per item for a column of unique values
+  const sortedGroups = useMemo(() => {
```
jwindhager: So we cannot sort descending as before, right? Just double-checking that this is intentional.

apps/tissuumaps/src/components/widgets/AnnotationsWidget/index.tsx
```diff
@@ -10,19 +8,26 @@ import { useTableColumnSelector } from "@/hooks/useTableColumnSelector";
...
 
 export type AnnotationsWidgetProps = {
   data?: ItemsData;
   tableHeight: number;
   table: string | null;
   selectedGroupByColumn?: string | null;
   onSelectedGroupByColumnChange?: (column: string | null) => void;
-  extraTableColumnDefs?: ColumnDef<AnnotationsTableRowData>[];
-  extraTableGroupColumnDefs?: ColumnDef<AnnotationsTableGroupRowData>[];
+  extraTableColumnDefs?: AnnotationsItemTableColumnDef[];
```
jwindhager: rename to something with "item", following the split?

jwindhager (changes requested):
Looks mostly good to me, thanks! Few comments below.

I'd rename to one of those (same for column defs; I'd tend towards the latter):
- `AnnotatedItemsTable` / `AnnotatedGroupsTable` (focus on the items/groups)
- `ItemAnnontationsTable` / `GroupAnnotationsTable` (focus on the annotations)

There seem to be different names for the same thing (e.g. height, content height, viewport height, rows height, ...). Or are these all different things?

I noticed that there are many useLayoutEffect hooks with empty dependency arrays ([]). This is generally considered bad practice and an anti-pattern in React, but I'm not sure this applies here? In general, useLayoutEffect should only be used when measuring DOM elements or adjusting their layout (which is definitely the case here) and consolidated into as few effects as possible to avoid layout trashing.

I did not use AI during review.

apps/tissuumaps/src/components/common/virtual-table.tsx
cavenel: Hmm not sure, this is a browser limitation, why would we allow a caller to change it?

apps/tissuumaps/src/components/common/virtual-table.tsx
cavenel: `compression`?

apps/tissuumaps/src/components/common/virtual-table.tsx
cavenel: Yes!

apps/tissuumaps/src/components/common/virtual-table.tsx
cavenel: See main comment

apps/tissuumaps/src/components/common/virtual-table.tsx
cavenel: See main comment

apps/tissuumaps/src/components/common/virtual-table.tsx
cavenel: See main comment

apps/tissuumaps/src/components/common/virtual-table.tsx
cavenel: Ok for renaming and moving to `hooks/useCompressedRowVirtualizer.ts`

apps/tissuumaps/src/components/widgets/AnnotationsWidget/GroupAnnotationsTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+        });
+    }
+    return () => {
+      abortController.abort();
+    };
+  }, [tableData, groupByColumn]);
+
+  // the groups are sorted as the plain values they are loaded as; a row object
+  // per group would cost as much as one per item for a column of unique values
+  const sortedGroups = useMemo(() => {
```
cavenel: We couldn't before, right? The sort will come in the groupby-table-preview PR later.

apps/tissuumaps/src/components/widgets/AnnotationsWidget/index.tsx
cavenel: Sure: `extraItemColumnDefs` `extraGroupColumnDefs`

cavenel:
> Looks mostly good to me, thanks! Few comments below.
> 
> I'd rename to one of those (same for column defs; I'd tend towards the latter):
> 
> * `AnnotatedItemsTable` / `AnnotatedGroupsTable` (focus on the items/groups)
> * `ItemAnnontationsTable` / `GroupAnnotationsTable` (focus on the annotations)

Yes sure, `ItemAnnotationsTable` / `GroupAnnotationsTable`

> There seem to be different names for the same thing (e.g. height, content height, viewport height, rows height, ...). Or are these all different things?

They are different things:
- containerHeight (the hook's parameter, the height prop of VirtualTable) is the container's CSS height.
- viewportHeight is the container's measured clientHeight. It excludes the horizontal scrollbar, and starts at the container height.
- contentHeight is the header plus all rows, uncompressed.
- layoutHeight is the height actually laid out once the content is compressed, and layoutRowsHeight is that height without the header.
I renamed them this way to make it clear, and compressScrollRange now talks about heights instead of sizes.

> I noticed that there are many useLayoutEffect hooks with empty dependency arrays ([]). This is generally considered bad practice and an anti-pattern in React, but I'm not sure this applies here? In general, useLayoutEffect should only be used when measuring DOM elements or adjusting their layout (which is definitely the case here) and consolidated into as few effects as possible to avoid layout trashing.

I think the empty deps are fine here. The effects only attach listeners to elements that mount once. Still there are many, so I merged the three into one, looks simpler!

apps/tissuumaps/src/components/common/virtual-table.tsx
jwindhager: Alright! Was thinking that we may not need the same max scroll height for all tables (making those "safer" by supporting potentially more browsers). But probably too niche to warrant making this a prop!

apps/tissuumaps/src/components/widgets/AnnotationsWidget/GroupAnnotationsTable.tsx
jwindhager: I somehow remember repeatedly clicking on the table header, but maybe my mind is playing tricks! Resolving

jwindhager (approved):

PR #187, opened 2026-09-12
feat(storage): multi-channel OME-Zarr images and OME-Zarr labels

jwindhager:

No directly related Jira issue.

Built on `refactor/labels` (#188, merged), which provides the `LabelsData` / `LabelsDataProvider` interfaces and `RasterData.getTileData()` used here, and on #191 (merged), which lets the renderer derive default contrast limits from a channel histogram reported by the image data. Rebased onto the current `development`.

- [omezarr-tilesource](https://github.com/TissUUmaps/omezarr-tilesource) v0.2.0 → v0.5.0 (`OMEZarrTileSource.open()`, shared `NgffImage`, `dataType: "ome-zarr"` raw chunk tiles)
- [ome-zarr.js](https://github.com/BioNGFF/ome-zarr.js) (`NgffImage`, OMERO channel metadata, `getSlices()`)
- [zarrita](https://github.com/manzt/zarrita.js) (`zarr.get()` for reading histogram planes) and [@zarrita/storage](https://github.com/manzt/zarrita.js) (`ZipFileStore` for zipped OME-Zarr)

# Type of change

- [ ] Other

Breaking: `sizeC` is removed from `OMEZarrImageDataSource` and its form. Projects that set it have the field ignored; the channel count is read from the OME-Zarr metadata.

# List of changes made

**Multi-channel images from metadata** (`OMEZarrImageDataProvider.ts`, `OMEZarrImageDataSource.ts`, `openOMEZarr.ts`)

- Store handling (remote store, `.ozx` URL, zipped workspace file) is factored into `openOMEZarr()`, shared by the image and labels providers; the earlier abstract base classes are gone.
- The provider loads the `NgffImage` once, pre-opens all resolution levels, and reads the channel count from the `c` axis of the full-resolution array. Images **with** a `c` axis are multi-channel, even if it has length one; images **without** a `c` axis are single-channel.
- One `OMEZarrTileSource` per channel via `OMEZarrTileSource.open(config, image)`, sharing the `NgffImage`; channels are opened concurrently, each followed by `signal?.throwIfAborted()` (`open()` does not throw on abort itself).
- Failures after the object URL is created revoke it before rethrowing.

**Channel histograms and default contrast limits** (`OMEZarrImageDataProvider.ts`, `OMEZarrImageData.ts`)

- As soon as a channel's tile source has opened, the provider reads the plane being shown (`c`, `z`, `t`, with `z`/`t` defaulting to the OMERO defaults like the tile source) from the lowest resolution level whose longer axis is at least 512 px, via `getSlices()` + `zarr.get()`. `MathUtils.computeRange()` gives the value range, `MathUtils.computeHistogram()` bins it (256 bins).
- The histograms are stored on `OMEZarrImageData` and returned by `getChannelHistogram(c)`, so the renderer can derive quantile-based default contrast limits (#191) for channels without an OMERO window instead of stretching over the data type range. 64-bit integer planes and planes without finite values yield `undefined`.

**Channel metadata and raw tiles** (`OMEZarrImageData.ts`)

- Single- and multi-channel images share one class: `getSizeC()` returns the channel count or `undefined`, `getTileSource(c?)` the per-channel or the single tile source.
- `getTileData()` returns the raw chunk of an `"ome-zarr"` tile as `{ values, width, height }`; 64-bit integer chunks are rejected, as their values cannot be represented in a `NumericArray` without loss. The renderer contrast-stretches and colorizes channels itself instead of receiving pre-rendered RGB tiles.
- `getChannelName()`, `getChannelVisibility()`, `getChannelColor()` and `getChannelContrastLimits()` read the OMERO `label`, `active`, `color` and `window`. Channels without a valid 6-digit hex color return `undefined` (also when other channels do have one), leaving the index-based default color to the renderer. Degenerate windows (`start >= end`) widen to `[start, start + 1]`.

**OME-Zarr labels** (`OMEZarrLabelsDataSource.ts`, `OMEZarrLabelsData.ts`, `OMEZarrLabelsDataProvider.ts`, `apps/tissuumaps/src/data/providers.ts`)

- `OMEZarrLabelsDataSource`: type `"ome-zarr"`, optional `z` / `t`, plus the inherited `table` reference.
- `OMEZarrLabelsDataProvider`: same store handling as the image provider via `openOMEZarr()`, one `OMEZarrTileSource` with `dataType: "ome-zarr"`. Schema exposes `url`, `z`, `t`, `table`. Registered in the app under `omeZarrLabelsDataSourceType` (the OpenSeadragon image provider is now registered before the OME-Zarr one).
- `OMEZarrLabelsData.getTileData()` returns label IDs as an unsigned array: Uint8/16/32 chunks pass through, any other data type throws. Label IDs are not enumerated (per #188, the renderer resolves them per tile). `image-label` metadata is not read.

**Zipped OME-Zarr** (both providers)

- Workspace paths are opened with `ZipFileStore.fromBlob(file)`, `.ozx` URLs with `ZipFileStore.fromUrl(url)`, other URLs as plain stores. For workspace files the object URL only serves as a unique tile cache key.

**Documentation**

- New docs page `supported-data/ome-zarr.md` (data source fields, channel settings and fallbacks, labels, example project, limitations); `development/dependencies.md` lists the OME-Zarr dependencies.

**Dependencies / build / housekeeping**

- `omezarr-tilesource` `^0.2.0` → `^0.5.0`; new direct deps `ome-zarr.js` `^0.0.20`, `@zarrita/storage` `^0.2.0` and `zarrita` `^0.7.5`, all externalized in the storage package's rolldown config. `omezarr-tilesource@0.5.0` is added to `minimumReleaseAgeExclude`.
- `OpenSeadragonImageData` aligned with the OME-Zarr class (explicit `undefined` checks, error message).

# Screenshot of the fix

N/A. The "Number of channels" field is gone from the OME-Zarr image form; OME-Zarr appears as a labels data source type.

# Use of AI

Claude Code was used during development and for this PR description.

# Testing

- `pnpm install` and rebuild `@tissuumaps/storage` (new dependencies).
- No automated tests: the storage package has no test suite and the repo has no OME-Zarr fixtures.
- The histogram path (plane selection, slicing, range and binning) was checked with a standalone Node script against public IDR images (`idr0062A/6001240.zarr`, 2 channels, uint16, 275 px; `idr0101A/13457227.zarr`, 5D, 4 channels, 2048 px): the expected resolution level is read as a 2D plane per channel and the derived default contrast limits are in the range of the images' curated OMERO windows.
- Manual checks:
  1. Multi-channel OME-Zarr with OMERO metadata: channel names, colors, visibility and contrast limits come from the metadata.
  2. Multi-channel OME-Zarr without OMERO windows: channels are stretched to histogram-derived limits rather than the full data type range.
  3. Single-channel or RGB OME-Zarr without a `c` axis: loads as one tile source as before; a `c` axis of length one loads as a one-channel multi-channel image.
  4. Zipped OME-Zarr via `.ozx` URL and via workspace file.
  5. OME-Zarr label image as a labels object: random palette, background `0` transparent. Cover a `uint16` array, and confirm `int32` / `int64` arrays fail loudly.
  6. Labels with `labelColor` from a referenced table: listed labels take table colors, unlisted labels the fallback.
  7. Load failure and mid-load abort for an image and a labels object: no leaked `blob:` URL, abort propagates.

# Further comments

- Loading a multi-channel image reads one downsampled plane per channel up front (the lowest level with a longer axis ≥ 512 px, a private constant of the provider); for images with many channels this adds a few chunk requests and some main-thread binning (yielding via `AsyncUtils.forEach`) to load time.
- Unzipped OME-Zarr directories in a workspace are still unsupported (`// TODO path` in both schemas).
- The labels provider does not reject arrays with a `c` axis; a multi-channel array renders channel 0.


jwindhager:
Still open for discussion:
- How to treat images with a **channel axis of size 1**? Currently, the `OMEZarrImageDataProvider` treats them as multi-channel with a single data channel. IMO, this is the cleanest approach for OME-Zarr, and probably also for other multi-channel image file formats such as TIFF.
- What **channel color** to provide for channels without color metadata? Currently, `OMEZarrImageData.getColor` returns `undefined` for such channels, even in the partial case where other channels do provide color information. This means that those channels get assigned the default color based on their index. *If this is the accepted way forward, the TSDoc comment on the `ImageData` interface must be updated before merging.*
- Which **channel contrast limits** should apply by default? Currently, `OMEZarrImageData.getChannelContrastLimits` provides contrast limits from file metadata, or `undefined`; automatic (e.g. quantile-based) contrast limits are not computed. *If we were to implement auto-contrast, we should do so generically on the viewer level, perhaps as optional feature - not only to ensure consistency across data providers, but also because data providers cannot access cached data.*

cavenel:
> How to treat images with a channel axis of size 1? 

I think multi-channel with a single data channel is correct, but I would then add a white color if none is given in metadata. Nobody wants to open a grayscale image in red tones.

> What channel color to provide for channels without color metadata? Currently, OMEZarrImageData.getColor returns undefined for such channels, even in the partial case where other channels do provide color information. This means that those channels get assigned the default color based on their index. If this is the accepted way forward, the TSDoc comment on the ImageData interface must be updated before merging.

Yes, that is what QuPath does so I would say it is ok and would update TSDoc accordingly.

> Which channel contrast limits should apply by default? Currently, OMEZarrImageData.getChannelContrastLimits provides contrast limits from file metadata, or undefined; automatic (e.g. quantile-based) contrast limits are not computed. If we were to implement auto-contrast, we should do so generically on the viewer level, perhaps as optional feature - not only to ensure consistency across data providers, but also because data providers cannot access cached data.

I definitely think we can not do without auto-contrast (else any 32 bit image would open black, and a user would think it is just not working). I wouldn't put it on the viewer level though, too late in the process, and if we use the tiles that are sent to the viewer, the contrast might change depending on what is loaded etc.

For the cache, I would say the Tiff side is covered (we use geotiff's cache). And for OME-Zarr it would be cached by the browser standard HTTP cache.

For where to put it, we could have a shared helper in tissuumaps/storage?

jwindhager:
> but I would then add a white color if none is given

For consistency, I think we should do this in the viewer, not in each data provider (multi-channel image case):
```
if (image.channels[c].color !== undefined) {
  return image.channels[c].color;
}
if (imageData.getColor(c) !== undefined) {
  return imageData.getColor(c)
}
if (imageData.getSizeC() === 1) {
  return white;
}
return ImageUtils.getDefaultChannelColor(c);
```

> Yes, that is what QuPath does so I would say it is ok and would update TSDoc accordingly.

Will do (in a separate PR)!

> we can not do without auto-contrast

Agreed, the only question is whether this should be default (i.e., upon opening the image) or manual (e.g. ImageJ's auto-contrast feature). I'd vote for default, like you, *but only if we also show the histogram* in the channels UI.

> I wouldn't put it on the viewer level though

Looked further into that. I wouldn't *extract* the data on the viewer level, but I would *compute* the contrast limits on the viewer level, using data (i.e., the histogram) provided by the data provider. Roughly as follows (for multi-channel images):
```
if (image.getChannelContrastLimits(c) !== undefined) {
  return image.getChannelContrastLimits(c);
}
const histogram = image.getHistogram();
return ImageUtils.getDefaultContrastLimits(histogram);
```

What exactly the histogram contains (which resolution, which crop, ...) would be decided by the data provider.

> For the cache

I think its fair to delegate caching of the values extracted for computing the histogram to the data provider, since only the data provider knows how the computation works.

> For where to put it, we could have a shared helper in tissuumaps/storage?

We shouldn't have any shared helpers in @tissuumaps/storage that we expect third-party storage providers to need. But with the above design (getContrastLimits/getHistogram are responsibility of data provider, getDefaultContrastLimits(histogram) is responsibility of the viewer) we wouldn't need an extra helper anymore / the extra helper would be getDefaultContrastLimits on the viewer level.

What do you think?

cavenel:
> What do you think?

I agree with everything! ❤️

jwindhager:
The discussed changes where implemented upstream

cavenel:
Right now `pnpm run build` fails because of `tsc -b apps/tissuumaps`: `LabelsDataWrapper.ts` line 24 says `number[] | UintArray`, it should say `number[] | IntArray | UintArray`, because [5fa3b2d ](https://github.com/TissUUmaps/TissUUmaps4/pull/187/changes/5fa3b2d4725e1245d5bf00c67a230a45ded8e130) widened LabelsData.getTileData in core. Right?

cavenel (approved):
Ready to merge!

PR #189, opened 2026-09-14
Fix/viewport fit user control patch

jwindhager:

None

# Type of change

- [ ] Other

# List of changes made

- Fit bounds in reset-size event rather than in updateBounds to make it work with multiple renderers (otherwise, renderers would only fit to their own dummy, forgetting about other renderers)

# Use of AI

Claude Code was used for implementation

# Testing

# Further comments


cavenel (approved):
Looks good!

PR #190, opened 2026-09-14
fix(renderer): user-controlled viewport detection

jwindhager:

Follow-up to #186.

# Type of change

- [ ] Other

# List of changes made

#186 detected user control of the viewport from raw input events (`pointerdown`, `wheel`, `keydown`) on the canvas and the navigator element. That over-triggers: those events also fire for gestures that leave the viewport where it is, e.g. drawing a shape on the overlay, which then wrongly suppressed refitting.

- Take viewport control from OpenSeadragon's own canvas/navigator gesture events (`canvas-drag`, `canvas-scroll`, `canvas-pinch`, `navigator-click`, `navigator-drag`, `navigator-scroll`) instead of raw DOM input events. The navigator raises its events on the viewer, so it no longer needs separate wiring.
- Handle keyboard navigation in the existing `canvas-key` handler, restricted to the keys OSD actually navigates with (arrows, WASD, `Equal`/`Minus`) and ignoring modified presses.

# Screenshot of the fix

N/A

# Use of AI

Claude Code was used during development.

# Testing

- Pan/zoom by drag, scroll, pinch, navigator click/drag and arrow/WASD/+/- keys should each stop the viewport from being refitted as data loads.
- Drawing on the overlay should *not* take viewport control; the viewport should still refit on world bounds changes.
- The home button and the `0` key binding should still hand control back.

# Further comments

None


cavenel (approved):
👍

PR #191, opened 2026-09-15
feat(renderer): histogram-based default contrast limits

jwindhager:

Follow-up to [this comment](https://github.com/TissUUmaps/TissUUmaps4/pull/187#issuecomment-5677229190) on #187. Implements the design proposed there: the viewer resolves channel colors (image settings → data provider → white for single-channel images → index-based default) and computes default contrast limits from a histogram supplied by the data provider (`ImageUtils.getDefaultContrastLimits`), so that no shared helper in `@tissuumaps/storage` is needed. Also includes the TSDoc update promised there, telling data providers to return `undefined` for channels without metadata colors.

# Type of change

- [ ] Other

# List of changes made

- `ImageData.getChannelHistogram?()`: new optional getter returning per-channel bin counts and the value range they span (first bin = lower bound, last bin = upper bound). No data provider implements it yet.
- `ImageUtils.getDefaultContrastLimits(histogram, qlow = 0.01, qhigh = 0.999)`: derives contrast limits from a histogram by clipping the fraction `qlow` of the values at the bottom and `1 - qhigh` at the top, scanning from both ends over non-empty bins only. Degenerate input (fewer than two bins, zero total count, zero-width range) and limits that would not be ascending return the range as is.
- `MathUtils.computeRange(values, { signal })`: async, computes the finite minimum and maximum of a numeric array, yielding to the event loop via `AsyncUtils.forEach` and rejecting on abort. Non-finite values are ignored; if no finite value is found, the empty range `[Infinity, -Infinity]` is returned so callers can check `vmin < vmax`.
- `MathUtils.computeHistogram(values, range, n = 256, { signal })`: async, computes a histogram over a *given* range (typically from `computeRange`), for data providers to use. Each value is counted in its nearest bin (bin `0` = lower bound, last bin = upper bound, matching the `getChannelHistogram` contract), values outside the range go to the edge bins, non-finite values are ignored, and a degenerate range puts everything into bin `0`. Yields and aborts like `computeRange`.
- `CSVTableData.loadValueRange` and `WebGLPointsRenderer._getObjectBounds` now use `MathUtils.computeRange` instead of hand-rolled min/max loops; the latter became async and takes the abort signal, matching the shapes renderer.
- `OpenSeadragonImageRenderer`: color and contrast-limit resolution moved into `_getChannelColor` / `_getChannelContrastLimits`. Contrast limits now fall back image settings → data → histogram → data type range. Single-channel images without a configured or reported color now default to white instead of the index-based red.
- `ImageDataWrapper` (app data cache) forwards `getChannelHistogram`, so the fallback reaches the renderer.
- TSDoc on `ImageData`, `ImageChannel` and the renderer updated for the new fallback chain; `getChannelColor` guidance changed to "return `undefined` for channels without metadata colors". Rendering docs page updated accordingly.
- Tests for `getDefaultContrastLimits` (clipping at both ends, default quantiles, quantiles 0 and 1, bin-to-bound mapping, collapsed and inverted limits, out-of-range quantile, degenerate input, round trip through `MathUtils.computeHistogram`), `computeRange` (plain and typed arrays, single value, non-finite values, empty range, large data, abort) and `computeHistogram` (nearest-bin assignment, bound-to-bin mapping, default bin count, out-of-range and non-finite values, degenerate range, single bin, empty data, large data, abort).

# Screenshot of the fix

Not applicable: no provider supplies a histogram yet, so rendering is unchanged except for the single-channel white default.

# Use of AI

Claude Code was used for development, review, TSDoc and tests.

# Testing

- `pnpm exec vitest run packages/@tissuumaps-core packages/@tissuumaps-storage packages/@tissuumaps-render`
- `pnpm exec tsc --noEmit` in core, storage and render, and `pnpm lint` pass.

# Further comments

- Out of scope, as discussed in the referenced comment: providing histograms from data providers (e.g. OME-Zarr in #187, which can now compute one from a downsampled level with `MathUtils.computeHistogram(values, await MathUtils.computeRange(values))`) and showing the histogram in the channels UI.
- Limits that would collapse or invert, e.g. when more than `qhigh - qlow` of the values fall into a single bin as for sparse fluorescence channels that are mostly background, fall back to the full histogram range, so channels never render black.
- `getChannelHistogram` is called per channel on every renderer synchronization, so implementations should return a precomputed histogram (noted in TSDoc).
- No sync variant of `computeRange` was added: the remaining min/max loops (shapes renderer bounds, rasterizer polygon extents, Parquet row-group statistics) operate on interleaved or non-array data and would not fit a flat-array helper.


cavenel (approved):
👍

PR #192, opened 2026-09-16
Tmap 194 add accessibility attributes to buttons without labels

KattisLej:

Closes [#TMAP-194](https://scilifelab.atlassian.net/browse/TMAP-194)

Based on [WCAG guidelines](https://webaim.org/standards/wcag/checklist) and  [W3 Accordion example](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/).

# Type of change

# List of changes made

- Checked the buttons with no text or that have an icon to add an aria-label
- Reverted the arrow direction to comply with the accordion example on W3, down up.

# Screenshot of the fix

image

# Use of AI

No.

# Testing

Make sure the changes make sense and that the accordion is working as the above example.


apps/tissuumaps/src/components/panels/ProjectPanel/index.tsx
```diff
@@ -99,6 +100,7 @@ export function ProjectPanel({ className }: ProjectPanelProps) {
         <Field>
           <Input
             ref={loadProjectFileInputRef}
+            aria-label="File failure message field"
```
cavenel, line 103: This label does not match the input, and the input is `hidden` so it is never read. We should remove it.

apps/tissuumaps/src/components/panels/ProjectPanel/index.tsx
```diff
@@ -70,11 +70,12 @@ export function ProjectPanel({ className }: ProjectPanelProps) {
     <div className={className}>
       <div>
         <Field>
-          <FieldLabel>Project name</FieldLabel>
+          <FieldLabel htmlFor="project-name">Project name</FieldLabel>
```
cavenel, line 73: I believe this is not needed (same for `id` on line 78). Base UI links `FieldLabel` and `FieldControl` on its own.

apps/tissuumaps/src/components/panels/PointsPanel/index.tsx
```diff
@@ -170,6 +171,7 @@ function PointsAccordionItem({ points, index }: PointsAccordionItemProps) {
             </Button>
             <Button
               variant="ghost"
+              aria-label="Delete points"
```
cavenel, line 174: When there is a title, aria-label and title should be the same.

apps/tissuumaps/src/components/panels/ShapesPanel/index.tsx
```diff
@@ -152,6 +153,7 @@ function ShapesAccordionItem({ shapes, index }: ShapesAccordionItemProps) {
             </Button>
             <Button
               variant="ghost"
+              aria-label="Delete"
```
cavenel, line 156: When there is a title, aria-label and title should be the same.

apps/tissuumaps/src/components/panels/TablesPanel/index.tsx
```diff
@@ -94,6 +94,7 @@ function TableAccordionItem({ table, index }: TableAccordionItemProps) {
           <div className="ml-auto flex flex-row items-center gap-x-2">
             <Button
               variant="ghost"
+              aria-label="Delete"
```
cavenel, line 97: When there is a title, aria-label and title should be the same.

apps/tissuumaps/src/components/panels/ProjectPanel/LayersWidget/index.tsx
```diff
@@ -162,6 +163,7 @@ function LayerAccordionItem({ layer, index }: LayerAccordionItemProps) {
             <Button
               variant="ghost"
               disabled={hasObjects}
+              aria-label="delete"
```
cavenel, line 166: This should be `aria-label="Delete layer"`. The dynamic title is still read as the description.

apps/tissuumaps/src/components/panels/ImagesPanel/index.tsx
```diff
@@ -122,6 +122,7 @@ function ImageAccordionItem({ image, index }: ImageAccordionItemProps) {
             </InputGroup>
             <Button
               variant="ghost"
+              aria-label="Show/hide"
```
cavenel, line 125: We should follow the state here: `aria-label={image.visibility ? "Hide image" : "Show image"}`

apps/tissuumaps/src/components/panels/LabelsPanel/index.tsx
```diff
@@ -141,6 +141,7 @@ function LabelsAccordionItem({ labels, index }: LabelsAccordionItemProps) {
             </InputGroup>
             <Button
               variant="ghost"
+              aria-label="Show/Hide"
```
cavenel, line 144: Same as for images: `aria-label={labels.visibility ? "Hide labels" : "Show labels"}`

apps/tissuumaps/src/components/panels/PointsPanel/index.tsx
```diff
@@ -162,6 +162,7 @@ function PointsAccordionItem({ points, index }: PointsAccordionItemProps) {
             </InputGroup>
             <Button
               variant="ghost"
+              aria-label="Show/hide"
```
cavenel, line 165: Same as for images: `aria-label={points.visibility ? "Hide points" : "Show points"}`

apps/tissuumaps/src/components/panels/ShapesPanel/index.tsx
```diff
@@ -144,6 +144,7 @@ function ShapesAccordionItem({ shapes, index }: ShapesAccordionItemProps) {
             </InputGroup>
             <Button
               variant="ghost"
+              aria-label="Show/hide"
```
cavenel, line 147: Same as for images: `aria-label={shapes.visibility ? "Hide shapes" : "Show shapes"}`

apps/tissuumaps/src/components/panels/ProjectPanel/LayersWidget/index.tsx
```diff
@@ -153,6 +153,7 @@ function LayerAccordionItem({ layer, index }: LayerAccordionItemProps) {
             </InputGroup>
             <Button
               variant="ghost"
+              aria-label="Show/hide"
```
cavenel, line 156: Same as for images: `aria-label={layer.visibility ? "Hide layer" : "Show layer"}`

apps/tissuumaps/src/components/common/accordion.tsx
```diff
@@ -76,8 +76,8 @@ export function AccordionTriggerUpDownIcon({
       className={cn("group/accordion-trigger", className)}
       {...props}
```
cavenel, line 77: This button has no name yet. We should add `aria-label="Expand/collapse"` before `{...props}`, here and in `AccordionTriggerRightDownIcon` above.

apps/tissuumaps/src/components/ui/dialog.tsx
```diff
@@ -63,6 +63,7 @@ function DialogContent({
             render={
               <Button
                 variant="ghost"
+                aria-label="Close"
```
cavenel, line 66: This duplicates the sr-only (screen-reader only) span on line 73. We should keep only one of the two.

cavenel (changes requested):
Thanks for taking this on. A few comments inline.

For the ones that already have a `title`: `aria-label` becomes the name and `title` becomes the description, and screen readers read both, so the two should say the same thing I think.

There are a few extra places that could benefit from aria-label also, maybe check for each one if you believe it can add value to have a correct description!

- **Dark mode toggle**, `apps/tissuumaps/src/App.tsx` line 89. Icon only, no name.
- **Opacity and point size inputs**, `ImagesPanel` line 106, `LabelsPanel` line 125, `PointsPanel` lines 128 and 146, `LayersWidget` line 137, `ShapesPanel` line 128. Their only label is the `α` or `s` in `InputGroupAddon`, which renders a `div` with `role="group"`, not a label. They are announced as an unnamed spin button.
- **Source select**, `DataSourceWidget/index.tsx` line 62. The `Source:` text is in a `FieldsetLegend` next to the select, not in a `Field`, so it is not the select's name.
- **Prompt input**, `dialogs/PromptDialog/index.tsx` line 75. No `Field` and no label. The dialog title gives the context visually but is not linked to the input.
- **Move up / move down / remove buttons**, `DataSourceWidget/controls/ArrayControl.tsx` lines 67, 73 and 79. No name.
- **Drag handles**, the six panels (`ImagesPanel` line 97 and the same in the others). `<GripVertical ref={handleRef} />` renders a bare `svg`. No name, no role, and not focusable, so reordering cannot be done with a keyboard at all. This one needs a `button` wrapper, not just a label.

apps/tissuumaps/src/components/panels/PointsPanel/index.tsx
KattisLej: Is this a particular request you have in the application? Or do you mean this as a general recommendation? In that case it would be great if you could provide me some reference to it. From what I can see an example from the[ W3](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/)(web accessibility initiative) the two fields are not identical: 

```
<label>Code
  <input name="code"
    aria-label="One-time code"
    placeholder="123456"
    title="Get your code from the app.">
</label>
```

I found [this article](https://www.a11y-collective.com/blog/aria-label-vs-title/) from A11 collective explaining it pretty well as the title being a visual tool for showing a tooltip on hovering where the aria label should be much more lessworded because it's main purpose is accessibility.

apps/tissuumaps/src/components/panels/ProjectPanel/index.tsx
KattisLej: If I remove it a get a "missing form label" error when using the WAVE evaluation tool. 

Image

So I would like to keep it but I could change the text of the aria label.

KattisLej: This is the tool I use:[ WAVE tools](https://wave.webaim.org/)

apps/tissuumaps/src/components/panels/ShapesPanel/index.tsx
KattisLej: I will change it to this to make it more specific.

apps/tissuumaps/src/components/panels/TablesPanel/index.tsx
KattisLej: Same here.

KattisLej:
@cavenel Thank you for your review. I have three new commits that I would like you to take a look at that are handling the most part of your suggestions. You can see what I have addressed in the review comments directly in the code. 

Also for the parts mentioned in your review comment:

_Dark mode toggle, apps/tissuumaps/src/App.tsx line 89. Icon only, no name._
- Fixed
_Opacity and point size inputs, ImagesPanel line 106, LabelsPanel line 125, PointsPanel lines 128 and 146, LayersWidget line 137, ShapesPanel line 128. Their only label is the α or s in InputGroupAddon, which renders a div with role="group", not a label. They are announced as an unnamed spin button._
- Fixed
_Source select, DataSourceWidget/index.tsx line 62. The Source: text is in a FieldsetLegend next to the select, not in a Field, so it is not the select's name._
- This I felt needed more than just adding a aria label, probably adjusting the component of simple-select to take a label, maybe as in the[ documentation](https://base-ui.com/react/components/select) a Select.Label (or SelectPrimitive.Label in this case)
_Prompt input, dialogs/PromptDialog/index.tsx line 75. No Field and no label. The dialog title gives the context visually but is not linked to the input._
- Fixed
_Move up / move down / remove buttons, DataSourceWidget/controls/ArrayControl.tsx lines 67, 73 and 79. No name._
- Fixed
_Drag handles, the six panels (ImagesPanel line 97 and the same in the others). <GripVertical ref={handleRef} /> renders a bare svg. No name, no role, and not focusable, so reordering cannot be done with a keyboard at all. This one needs a button wrapper, not just a label._
- This also sounds like a separate issue, as you suggested it's not only a label. 

Also I am not sure how to handle the Codecov 91.11% coverage fail.

apps/tissuumaps/src/components/widgets/DataSourceWidget/index.tsx
```diff
@@ -58,7 +58,7 @@ export function DataSourceWidget<TDataSource extends DataSource>({
       <FieldsetLegend className="flex flex-row items-center gap-x-1 font-medium text-foreground">
         {isEditing ? (
           <>
-            Source:
+            <FieldLabel>Source</FieldLabel>
```
cavenel: `FieldLabel` is not imported, so this does not compile: `tsc -b apps/tissuumaps/tsconfig.json` gives `TS2304: Cannot find name 'FieldLabel'`. CI does not typecheck, which is why lint and test are green.

Since this one needs a proper `Field` / `Select.Label` anyway, please revert this line back to `Source:` and handle it in the follow-up issue.

apps/tissuumaps/src/components/panels/ProjectPanel/index.tsx
```diff
@@ -99,6 +99,7 @@ export function ProjectPanel({ className }: ProjectPanelProps) {
         <Field>
           <Input
             ref={loadProjectFileInputRef}
+            aria-label="File failure message field"
```
cavenel: "File failure message field" does not describe what the control does. `aria-label="Load project from file"`.

apps/tissuumaps/src/App.tsx
```diff
@@ -86,7 +86,9 @@ function DockviewRightHeaderActionsComponent() {
   const dark = useSettingsStore((state) => state.dark);
   const setDark = useSettingsStore((state) => state.setDark);
   return (
-    <Button onClick={() => setDark(!dark)}>{dark ? <Sun /> : <Moon />}</Button>
+    <Button aria-label="Dark mode toggle" onClick={() => setDark(!dark)}>
+      {dark ? <Sun /> : <Moon />}{" "}
```
cavenel: Stray `{" "}` after the icon, please remove.

apps/tissuumaps/src/components/panels/LabelsPanel/index.tsx
cavenel: This one was missed I think?

apps/tissuumaps/src/components/panels/PointsPanel/index.tsx
cavenel: Yes you are right, it's fine to have different values!
I think your changes now are perfect ("Delete table", "Delete shape cloud").
Maybe just replace "Delete points" with "Delete point cloud", like you did in ShapesPanel?

apps/tissuumaps/src/components/panels/ProjectPanel/index.tsx
cavenel: Ok to keep for the hidden field then!
But "File failure message field" does not describe what the control does. `aria-label="Load project from file"` would be a better fit.

apps/tissuumaps/src/components/widgets/DataSourceWidget/index.tsx
KattisLej: Oh, this was surely a miss from my side, sorry. My intention was to leave it as is since I didn't fully have the understanding of Base UI components. One idea would be to adjust the component of simple-select to take a label, but your suggestion seems also very interesting and simpler. 
- Should I make the change you proposed? (and of course import the label)

apps/tissuumaps/src/App.tsx
KattisLej: This is actually a space added by my editor. I will remove it.

apps/tissuumaps/src/components/widgets/DataSourceWidget/index.tsx
KattisLej: I made a commit with the change you proposed, let me know what you think.

apps/tissuumaps/src/components/widgets/DataSourceWidget/index.tsx
```diff
@@ -58,21 +59,23 @@ export function DataSourceWidget<TDataSource extends DataSource>({
       <FieldsetLegend className="flex flex-row items-center gap-x-1 font-medium text-foreground">
         {isEditing ? (
           <>
```
cavenel: The <>...</> can now be removed, as it contains a single child

cavenel (approved):
One tiny final detail and then you can merge! Thanks a lot!!

apps/tissuumaps/src/components/widgets/DataSourceWidget/index.tsx
KattisLej: Yes, thank you for this.

PR #193, opened 2026-09-16
feat(core): seeded RandomUtils, FNV-1a/fmix32 HashUtils and histogram subsampling

jwindhager:

No linked issue. Algorithms referenced in the code:

- FNV-1a: http://www.isthe.com/chongo/tech/comp/fnv/
- MurmurHash3 `fmix32` finalizer: https://github.com/aappleby/smhasher/blob/master/src/MurmurHash3.cpp
- SplitMix32 (JS constants): https://github.com/bryc/code/blob/master/jshash/PRNGs.md#splitmix32

# Type of change

- [ ] Other

Breaking for consumers of `@tissuumaps/core`: the public `HashUtils` API changed (see below) and `MathUtils.computeHistogram` now defaults to 1024 bins instead of 256. Within this repo all callers are updated.

# List of changes made

**`HashUtils` (core) rebuilt on FNV-1a and the MurmurHash3 finalizer**

- Removed `djb2`, `djb2Pick`, `lowbias32` and `lowbias32Pick`.
- Added a two-step API: `hashRaw(key)` reduces a string to a 32-bit FNV-1a hash (over UTF-16 code units, so non-Latin-1 text differs from byte-oriented implementations), `mix(h, seed)` XORs in a seed and runs the `fmix32` avalanche step, and `hash(key, seed)` does both. Callers that hash the same key with many seeds can cache the raw hash.
- Exposed the primitives `fnv1a(str)` and `fmix32(h, config)`; the finalizer constants are parametrized through the new exported `Fmix32Config` type so other algorithms can reuse the same code path (SplitMix32 does, see below).
- Palette picking moved to the call sites as `palette[HashUtils.hash(group) % palette.length]` (app annotation columns for labels, points and shapes; `ColorResolver` and `MarkerResolver` in the renderer). `ColorResolver.pickRandomColor` uses `HashUtils.mix(id, seed) % palette.length`.
- Tests rewritten against known FNV-1a and MurmurHash3 vectors, plus injectivity and scattering checks.

**New `RandomUtils` (core): seeded, deterministic PRNG utilities**

- `splitmix32(seed)` / `createUint32RNG(seed)`: a SplitMix32 generator (Weyl counter with increment `0x9e3779b9`, output through `fmix32` with the SplitMix32 constants). Period 2^32, every 32-bit value once per period.
- `randUint32(seedOrRng)` and `randUnitFloat32(seedOrRng)`: first draw for a seed, or next draw of an existing generator.
- `randUint32Array(n, seedOrRng, array?)` and `randUnitFloat32Array(n, seedOrRng, array?)`: fill the first `n` elements of a new or given typed array; throw `RangeError` if the given array is too short.
- `toUnitFloat32(uint32)`: maps the top 24 bits to `[0, 1)` so values are exact in a `Float32Array`.
- `seed()`: the one non-deterministic method, a `Math.random()`-based 32-bit seed for "shuffle" actions.
- Exported from the package index. Tests use reference vectors from an independent SplitMix32 implementation, and cover determinism, prefix behavior, in-place fills and the float exactness guarantee.

**`MathUtils.computeHistogram` (core): optional seeded subsampling**

- Parameter `n` renamed to `bins`; default changed from 256 to 1024.
- New options `sample` and `seed`. A positive `sample` below the number of values estimates the histogram from that many values drawn uniformly with replacement via `RandomUtils.createUint32RNG(seed)`. Omitting it, `0`, or a value at or above the length counts every value once.
- Implemented as a single `AsyncUtils.forEach` pass over a length-only array-like, reading values by sequential or random index. No intermediate index or sample arrays are allocated, so memory is constant regardless of `sample`.
- Tests added for sample count, determinism per seed, whole-array coverage, correct binning of fractional and negative values under sampling, non-finite handling, the full-pass fallback and abort.

**App**

- The "Shuffle seed" button in `ColorConfigWidget` now calls `RandomUtils.seed()`.

**Fix included in this branch**

- The last commit fixes `ColorResolver.pickRandomColor`, which after the `HashUtils` rewrite indexed the palette with the raw 32-bit hash and ignored the seed, returning `undefined`. Nine renderer tests caught this; they pass again.

# Screenshot of the fix

Not applicable.

# Use of AI

Claude Code was used during development: for TSDoc, for the test suites of `HashUtils`, `RandomUtils` and `computeHistogram`, for the single-pass sampling implementation of `computeHistogram`, and for reviewing the branch, which surfaced the `pickRandomColor` regression.

# Testing

- No other branches, new dependencies or configuration are needed.
- Run `pnpm -r test`. On this branch: core 335, renderer 226, app 87 tests pass.
- In the app, set a points layer to a random color config and press "Shuffle seed" repeatedly; each press should yield a new seed and a new color assignment. Reload and confirm the same seed reproduces the same colors.
- Load an image and check that the histogram-based default contrast limits still look right with the new 1024-bin default.

# Further comments

- The seed-or-generator parameter of the `rand*` methods lets callers continue one stream across several calls, e.g. `randUint32Array(k, rng)` twice equals `randUint32Array(2k, seed)` when `rng = createUint32RNG(seed)`.
- `computeHistogram` uses `rng() % length` rather than a unit float times the length: the 24-bit float cannot address every index of arrays above 16M elements, while the modulo bias of a 32-bit draw is negligible.


cavenel (approved):
👍

PR #194, opened 2026-09-16
Feat/geoparquet shapes

cavenel:

N/A

# Type of change

- [ ] Other

# List of changes made

- **GeoParquet shapes data provider** (`type: "parquet"` for shapes): reads the WKB polygon / multi-polygon column of a GeoParquet file in the Parquet worker, with `geometryColumn`, `idColumn`, `nameColumn` and `table` fields. Registered as a built-in provider.
- **Point geometries as table columns**: a GeoParquet point column `geometry` is exposed by the Parquet table provider as `geometry.x` / `geometry.y`, so circles can be drawn as a table point cloud (and sized by a sibling `radius` column). Their value range comes from the `geo` metadata `bbox`, without decoding the column.
- `GeoParquetUtils`: reads the `geo` metadata (WKB columns only, 2D and 3D bboxes) and resolves the derived coordinate columns. Unit-tested.
- `ShapesGeometryBuilder` (new `common/`): the CSR geometry accumulator moved out of the GeoJSON worker and shared with the Parquet worker. Unit-tested.
- **Breaking**: `GeoJSONShapesData` and `ParquetShapesData` are replaced by one `InMemoryShapesData` exported from `@tissuumaps/storage`.
- Docs: `supported-data/parquet.md` written, provider table updated.

# Screenshot of the fix

# Use of AI

Claude was used during development, reviewed by a human.

# Testing

- Open a project pointing at a SpatialData store, e.g. the `project.tmap` example in `apps/docs/docs/supported-data/parquet.md`: the circles element loads as a table with `geometry.x`/`geometry.y` columns and draws as a point cloud, the polygons element draws as shapes.
- `pnpm exec vitest run --root packages/@tissuumaps-storage` (18 tests).

# Further comments

- Only WKB-encoded geometry columns are handled; GeoParquet 1.1 native `point` encoding is read as raw values, like any non-geometry column.
- A real column named `<geometryColumn>.x` would be shadowed by the derived one; not guarded, documented behaviour.


packages/@tissuumaps-storage/src/parquet/parquet.worker.ts
```diff
@@ -298,6 +422,84 @@ async function handleColumnRequest(
...
+    );
+  }
+  if (GeoParquetUtils.isPointColumn(geoColumn)) {
+    throw new Error(
+      `Geometry column "${geoColumn.name}" holds points, which are read as ` +
+        `the "${geoColumn.name}.x" and "${geoColumn.name}.y" columns of a table`,
+    );
+  }
+  // Progress only tracks the geometry, which dwarfs the ID and name columns
+  const { ids: rowIds, names: rowNames } = await readIdsAndNames(
```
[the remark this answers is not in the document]
cavenel: Intentional. A missing or non-integer ID is a broken file, not a row to skip; the table provider rejects it the same way.

packages/@tissuumaps-storage/src/parquet/parquet.worker.ts
```diff
@@ -298,6 +422,84 @@ async function handleColumnRequest(
...
+  const geometry = builder.build();
+  if (geometry.shapePolygonOffsets.length === 1) {
+    throw new Error(`No valid geometries found in column "${geoColumn.name}"`);
+  }
+  return {
+    response: {
+      op: "shapes",
+      geometry,
+      ids: rowIds !== undefined ? ids : undefined,
+      names: rowNames !== undefined ? names : undefined,
```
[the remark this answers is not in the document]
cavenel: Fixed

packages/@tissuumaps-storage/src/parquet/GeoParquetShapesDataProvider.ts
```diff
@@ -0,0 +1,115 @@
...
+        type: "string",
+      },
+      nameColumn: {
+        type: "string",
+      },
+      table: {
+        type: "string",
+      },
+    },
+    required: ["url"], // TODO ... or path
```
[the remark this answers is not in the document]
cavenel: Pre-existing TODO shared by every provider. Will fix in an other PR.

apps/docs/docs/supported-data/parquet.md
```diff
@@ -3,3 +3,89 @@ sidebar_position: 7
...
+| ------------ | -------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| `type`       | `string` | Always `"parquet"`.                                                                                                            |
+| `url`        | `string` | URL of a remote Parquet file, absolute or relative (see [Referencing data](../concepts/projects.md#referencing-data)).         |
+| `path`       | `string` | Path of a Parquet file relative to the workspace directory (see [Referencing data](../concepts/projects.md#referencing-data)). |
+| `idColumn`   | `string` | Column holding the ID of each row (see [Data model](../concepts/data-model.md)). Row numbers are used when it is not given.    |
+| `nameColumn` | `string` | Column holding the name of each row.                                                                                           |
+
+Either `url` or `path` has to be given. When a workspace is open and both are given, `path` takes precedence.
+
+Columns of 64-bit integers are not supported, as their values do not fit a JavaScript number.
```
jwindhager: Slightly inaccurate (JS has support for bigint, but TissUUmaps doesn't as of now)

jwindhager (changes requested):
Before I review in-depth, as discussed offline:

Every data provider/type/etc should be named after its most general entity. For shapes, this would be "geoparquet" (and not "parquet" like it is today).

Instead of the flat geojson/parquet/common structure, use `geo/geojson` and `geo/geoparquet`, and put the shared code (ShapesGeometryBuilder, InMemoryShapesData, maybe GeoParquetUtils?),  into `geo`; the parquet table data provider should stay separate (same level as `geo`). If the GeoParquetShapesDataProvider really depends on code from the ParquetTableDataProvider, that's not a problem (use a relative `../parquet/...` import).

apps/docs/docs/supported-data/parquet.md
cavenel: Fixed!

cavenel:
Summary of what we settled, partly offline with @jwindhager and partly while building this out.

**Type string: `"parquet"` for both, not `"geoparquet"`.**

Same reasoning as TIFF. `qptiff` and `ome-tiff` are both read under `tiff`. A format variant does not get its own type string. The registries are per data kind (`shapesDataProviders`, `tableDataProviders`), so there is nothing to disambiguate.

**Structure: folders are per file format, shared code goes to core.**

No `common/`, no `geo/`.

```
@tissuumaps/core
└ src/storage/shapes.ts      unchanged, still only types
│  ├ type ShapesGeometry     CSR typed arrays, transferable
│  └ type ShapesData         getIds / getSize / loadGeometry
└ src/utils/ShapesUtils.ts   buildGeometry(append => ...)
                             createShapesData(geometry, ids, names)

@tissuumaps/storage
├ geojson/
│ ├ GeoJSONShapesDataProvider.ts
│ ├ GeoJSONShapesDataSource.ts
│ └ geojson.worker.ts         GeoJSON to ShapesUtils.buildGeometry
├ parquet/
│ ├ ParquetTableDataProvider.ts
│ ├ ParquetTableData.ts
│ ├ GeoParquetShapesDataProvider.ts
│ ├ GeoParquetUtils.ts        reads the `geo` metadata
│ └ parquet.worker.ts         WKB to ShapesUtils.buildGeometry, plus plain tables
└ csv/  ome-zarr/  openseadragon/  table/     unchanged
```

`geo/` does not work here because `parquet/` is not only geo. The table provider and the shapes provider share one worker and one file reader. A `geo/` folder would either split the format across two folders, or split the worker into two bundles that open the same file twice. So the folder is the file format, and anything shared across formats goes to core instead.

The builder needs state, but only for the duration of one build. That state lives inside the callback, so it fits as one static `ShapesUtils.buildGeometry(append => ...)` rather than a class. It also collects the shape IDs and names, so they cannot drift out of alignment with the geometry.

On `InMemoryShapesData`: agreed, we do not need a class. The behaviour was still there twice though, as two identical forty-line holders, one per format. So it went the same way as the builder. `ShapesUtils.createShapesData(geometry, ids, names)` returns the `ShapesData`. No class, no duplication, and nothing shared left inside `storage`.

`GeoParquetUtils` stays in `parquet/`. `ParquetTableData` uses it too, to expose point geometry columns as `geometry[x]` and `geometry[y]`.

**Room for GeoArrow later.**

Today we only decode geometry columns whose `geo` metadata says `encoding: "WKB"`. Columns in other encodings are read as raw values.

GeoParquet 1.1 also allows native encodings, which are GeoArrow nested lists instead of WKB. SpatialData can write geometry that way.

Adding support later stays inside `parquet/`. It means widening the filter in `GeoParquetUtils` and adding a decode path in `parquet.worker.ts`. No new folder and no new type string, because a GeoArrow encoded file is still a Parquet file.

apps/docs/docs/supported-data/parquet.md
```diff
@@ -3,3 +3,91 @@ sidebar_position: 7
...
+| `nameColumn`     | `string` | Column holding the name of each row.                                                                                           |
+| `requestHeaders` | `object` | Extra HTTP headers sent with the request for a remote file.                                                                    |
+
+Either `url` or `path` has to be given. When a workspace is open and both are given, `path` takes precedence.
+
+Columns of 64-bit integers are not supported, as TissUUmaps does not handle bigint values.
+
+## Shapes data source
+
+GeoParquet shapes data sources have the `type` `"parquet"` and accept the following fields:
```
jwindhager, line 28: I'd say either:


```suggestion
Parquet shapes data sources have the `type` `"parquet"` and accept the following fields:
```

or:


```suggestion
GeoParquet shapes data sources have the `type` `"geoparquet"` and accept the following fields:
```

I think [we discussed to do the former](https://github.com/TissUUmaps/TissUUmaps4/pull/194#issuecomment-5749107324) and treat GeoParquet as a variant of Parquet.

Same applies elsewhere in this file, e.g. in the first paragraph.

apps/docs/docs/supported-data/parquet.md
```diff
@@ -3,3 +3,91 @@ sidebar_position: 7
...
+| `idColumn`       | `string` | Column holding the ID of each shape. Row numbers are used when it is not given.      |
+| `nameColumn`     | `string` | Column holding the name of each shape.                                               |
+| `requestHeaders` | `object` | Extra HTTP headers sent with the request for a remote file.                          |
+| `table`          | `string` | ID of the table annotating the shapes (see [Data model](../concepts/data-model.md)). |
+
+Polygons and multi-polygons are read as shapes; rows holding another geometry are skipped.
+
+## GeoParquet
+
+A GeoParquet file describes its geometry columns in the `geo` metadata of the file, and stores their geometries as [WKB](https://libgeos.org/specifications/wkb/). Geometry columns in other encodings are read as their raw values.
```
jwindhager: > Geometry columns in other encodings are read as their raw values.

By the GeoParquetShapesDataSource? How would this work?

apps/docs/docs/supported-data/parquet.md
```diff
@@ -3,3 +3,91 @@ sidebar_position: 7
...
+| `path`           | `string` | Path of a GeoParquet file relative to the workspace directory.                       |
+| `geometryColumn` | `string` | Geometry column to read. Defaults to the primary geometry column of the file.        |
+| `idColumn`       | `string` | Column holding the ID of each shape. Row numbers are used when it is not given.      |
+| `nameColumn`     | `string` | Column holding the name of each shape.                                               |
+| `requestHeaders` | `object` | Extra HTTP headers sent with the request for a remote file.                          |
+| `table`          | `string` | ID of the table annotating the shapes (see [Data model](../concepts/data-model.md)). |
+
+Polygons and multi-polygons are read as shapes; rows holding another geometry are skipped.
+
+## GeoParquet
```
jwindhager, line 43: I think it would be useful to split this section and move it as subsection below the respective data provider (e.g. explain how GeoJSON points are represented in the table data provider, explain how shapes are encoded using GeoJSON in the shapes data provider, and refer from the shapes provider to the points provider for GeoJSON points)

apps/tissuumaps/src/data/providers.ts
```diff
@@ -1,20 +1,20 @@
 import {
   CSVTableDataProvider,
   GeoJSONShapesDataProvider,
+  GeoParquetShapesDataProvider,
```
jwindhager, line 4: Shouldn't this be ParquetShapesDataProvider, parquetShapesDataSourceType, etc now, following the decision that GeoParquet is just a variant of Parquet (similar to TIFF)?

packages/@tissuumaps-core/src/utils/ShapesUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Creates shapes data from a geometry that is already fully loaded
+   *
+   * @param geometry - The geometry of the shapes
+   * @param ids - The ID of every shape, in geometry order
+   * @param names - The name of every shape, if the shapes have names
+   * @returns The shapes data, holding the geometry in memory
+   * @throws Error if the IDs do not match the geometry
+   */
+  static createShapesData(
```
jwindhager, line 111: For consistency, I would either switch all data providers to anonymous objects (where possible), or use named GeoJSONShapesData and GeoParquetShapesData classes (I think that would be my preference, in case we add format-specific stuff in the future).

packages/@tissuumaps-core/src/utils/ShapesUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+   * only if every appended shape was given one.
+   *
+   * @param appendShapes - Callback appending the shapes, receiving the appender
+   * @returns The shapes geometry, in typed arrays ready to be transferred,
+   * along with the IDs and names of the appended shapes
+   */
+  // No signal option: this performs no cancellable work of its own, and
+  // appendShapes owns whatever it awaits.
+  static async buildGeometry(
+    appendShapes: (append: ShapesAppender) => void | Promise<void>,
```
jwindhager, line 46: Ok, sorry for suggesting earlier to have this as a single function - I realize now that this reads strictly worse than a dedicated builder class 😬 Just not sure where such a class would live in our project layout...

If you don't want to switch back to the builder class, I understand. I this case, how about renaming as follows:

```suggestion
    walk: (addShape: ...) => void | Promise<void>,
```

packages/@tissuumaps-storage/src/parquet/GeoParquetUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+      bbox?: number[];
+    };
+  };
+};
+
+/** Matches a coordinate column, capturing its geometry column and its axis */
+const coordinateColumnPattern = /^(.*)\[([xy])\]$/;
+
+/** Reads the 2D bounds of a `bbox`, which lists Z bounds too for 3D columns */
+function readBBox(
```
jwindhager, line 39: Any reason this (and coordinateColumnPattern) are not static class members?

packages/@tissuumaps-storage/src/parquet/ParquetTableData.ts
```diff
@@ -62,13 +70,43 @@ export class ParquetTableData implements TableData {
...
     const { data } = await runParquetWorker(
       { op: "column", source: this._source, column },
       { signal, onProgress },
     );
     return data as GenericArray<T>;
   }
 
+  // No signal option: the read is shared by both axes, so one caller must not
+  // be able to abort it for the other. Callers race it against their own
+  // signal instead, and the progress goes to whoever starts the read.
```
jwindhager, line 92: Why does this get called twice ("both axes")? If deduplication is really needed here by design, we should probably consider moving SharedOperation.ts into core...

jwindhager (commented):
Looks mostly good to me, thanks a lot!

I think the decision to treat GeoParquet as a "variant"/"profile" of Parquet (similar to TIFF) makes a lot of sense. Currently, however, this is only reflected in the data source `type`; file names etc still need updating (see my comment).

apps/docs/docs/supported-data/parquet.md
cavenel: Ok or "Parquet shapes data sources have the `type` `"parquet"` and accept the following fields"

apps/docs/docs/supported-data/parquet.md
cavenel: Sure!

apps/tissuumaps/src/data/providers.ts
cavenel: 👍

packages/@tissuumaps-core/src/utils/ShapesUtils.ts
cavenel: Ok for duplicated GeoJSONShapesData and GeoParquetShapesData!

packages/@tissuumaps-storage/src/parquet/ParquetTableData.ts
cavenel: A point cloud requests geometry[x] and geometry[y] as two separate column reads, so the shared promise stops the geometry column being decoded twice. There is no SharedOperation on this branch to move into core.

packages/@tissuumaps-storage/src/parquet/ParquetMetadataUtils.ts
```diff
@@ -0,0 +1,201 @@
...
+    const geo = metadata.key_value_metadata?.find(({ key }) => key === "geo");
+    if (geo?.value === undefined) {
+      return [];
+    }
+    const { primary_column, columns = {} } = JSON.parse(
+      geo.value,
+    ) as GeoMetadata;
+    return Object.entries(columns)
+      .filter(([, column]) => column.encoding === "WKB")
+      .map(([name, column]) => ({
```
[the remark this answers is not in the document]
cavenel: Correct, and deliberate for now. Will fix when we add GeoArrow support.

packages/@tissuumaps-storage/src/parquet/ParquetShapesData.ts
```diff
@@ -0,0 +1,39 @@
+import type { ShapesData, ShapesGeometry } from "@tissuumaps/core";
+
+/** Shapes data of a GeoParquet file, fully loaded up front */
+export class ParquetShapesData implements ShapesData {
```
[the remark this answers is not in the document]
cavenel: The description has been rewritten to the implemented design.

apps/docs/docs/supported-data/parquet.md
```diff
@@ -3,3 +3,107 @@ sidebar_position: 7
 ---
 
 # Parquet
+
+The built-in **Parquet data provider** opens Parquet files as **tables**, and [GeoParquet](https://geoparquet.org/) files as **shapes**. Files are read in the browser with [hyparquet](https://hyparquet.com/), one column at a time, so a remote file needs a server that supports HTTP range requests.
```
jwindhager:
```suggestion
The **Parquet data providers** open Parquet files as **tables**, and [GeoParquet](https://geoparquet.org/) files as **shapes**. Files are read in the browser with [hyparquet](https://hyparquet.com/), one column at a time, so a remote file needs a server that supports HTTP range requests.
```

apps/docs/docs/supported-data/parquet.md
```diff
@@ -3,3 +3,107 @@ sidebar_position: 7
...
+| ---------------- | -------- | -------------------------------------------------------------------------------------------------------- |
+| `type`           | `string` | Always `"parquet"`.                                                                                      |
+| `source`         | `string` | URL or path of the GeoParquet file (see [Referencing data](../concepts/projects.md#referencing-data)).   |
+| `geometryColumn` | `string` | Geometry column to read. Defaults to the primary geometry column of the file.                            |
+| `idColumn`       | `string` | Column holding the ID of each shape. Defaults to the [pandas index](#pandas-index), else to row numbers. |
+| `nameColumn`     | `string` | Column holding the name of each shape.                                                                   |
+| `requestHeaders` | `object` | Extra HTTP headers sent with the request for a remote file.                                              |
+| `table`          | `string` | ID of the table annotating the shapes (see [Data model](../concepts/data-model.md)).                     |
```
jwindhager: Why do points not have a table (see above)?

apps/docs/docs/supported-data/parquet.md
```diff
@@ -3,3 +3,107 @@ sidebar_position: 7
...
+
+### GeoParquet geometries
+
+A [GeoParquet](https://geoparquet.org/) file describes its geometry columns in the `geo` metadata of the file, and stores their geometries as [WKB](https://libgeos.org/specifications/wkb/). Geometry columns in other encodings are read as their raw values. A column holding points is not read as shapes but as a pair of columns of the table data source, see [Point geometries as coordinate columns](#point-geometries-as-coordinate-columns).
+
+## Pandas index
+
+Parquet has no index, so a file written by [pandas](https://pandas.pydata.org/docs/development/developer.html) or [GeoPandas](https://geopandas.org/) records in its `pandas` metadata which column its DataFrame index was written as. That column is the default `idColumn`, so that a [SpatialData](https://spatialdata.scverse.org/) element, whose index is the key its tables refer to, is keyed without configuration.
+
+Only a single-level integer index is used: IDs are numbers, so a string index leaves the default at row numbers, as do a multi-level index and a `RangeIndex`, which is not written as a column at all.
```
jwindhager: Is this describing how pandas/geopandas does it, or a limitation on our side?

apps/docs/docs/supported-data/parquet.md
```diff
@@ -3,3 +3,107 @@ sidebar_position: 7
...
+| `requestHeaders` | `object` | Extra HTTP headers sent with the request for a remote file.                                              |
+| `table`          | `string` | ID of the table annotating the shapes (see [Data model](../concepts/data-model.md)).                     |
+
+Polygons and multi-polygons are read as shapes; rows holding another geometry are skipped.
+
+### GeoParquet geometries
+
+A [GeoParquet](https://geoparquet.org/) file describes its geometry columns in the `geo` metadata of the file, and stores their geometries as [WKB](https://libgeos.org/specifications/wkb/). Geometry columns in other encodings are read as their raw values. A column holding points is not read as shapes but as a pair of columns of the table data source, see [Point geometries as coordinate columns](#point-geometries-as-coordinate-columns).
+
+## Pandas index
```
jwindhager: I would move this under a shared `## Supported flavors` section that contains subsection for:
- Vanilla
- Pandas (mention the index stuff)
- GeoParquet (perhaps move back the "provider paragraphs"?)
- SpatialData (mention that it builds on Pandas & GeoParquet with WKB)

apps/docs/docs/supported-data/parquet.md
```diff
@@ -3,3 +3,107 @@ sidebar_position: 7
...
+
+Parquet has no index, so a file written by [pandas](https://pandas.pydata.org/docs/development/developer.html) or [GeoPandas](https://geopandas.org/) records in its `pandas` metadata which column its DataFrame index was written as. That column is the default `idColumn`, so that a [SpatialData](https://spatialdata.scverse.org/) element, whose index is the key its tables refer to, is keyed without configuration.
```
jwindhager: Can we move "everything SpatialData" (e.g. that it must use WKB and a geopandas index in order to be supported by TissUUmaps) into a separate section, and keep this only about pandas?

packages/@tissuumaps-storage/src/parquet/parquet.worker.ts
```diff
@@ -8,10 +9,53 @@ import {
...
+  }
+  if (geometry.type === "MultiPolygon") {
+    builder.addShape(geometry.coordinates, id, name);
+    return;
+  }
+  console.warn(`Unsupported geometry type: ${geometry.type}`);
+}
+
+/** Why a point geometry column cannot be read as shapes */
+function pointColumnMessage(column: string): string {
```
jwindhager: this feels like too much abstraction for me

packages/@tissuumaps-storage/src/parquet/ParquetMetadataUtils.ts
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * The `pandas` metadata of a Parquet file written by pandas
+ *
+ * `index_columns` lists where the DataFrame index went: an entry is the name
+ * of the column it was written as, or a description of a `RangeIndex` that was
+ * not written at all. `columns` describes every column, the index ones
+ * included, and gives the pandas dtype each was written from.
+ */
+type PandasMetadata = {
```
jwindhager: Should this be in the ParquetMetadataUtils? Feels like it would warrant a separate PandasMetadataUtils?

packages/@tissuumaps-storage/src/parquet/ParquetTableData.ts
```diff
@@ -63,13 +81,41 @@ export class ParquetTableData implements TableData {
   ): Promise<GenericArray<T>> {
     const { signal, onProgress } = options ?? {};
     signal?.throwIfAborted();
+    const coordinates = this._coordinateColumns.get(column);
```
jwindhager: The `coordinates` variable name here is misleading, especially compared to `coordinates` in `_loadCoordinates`

jwindhager (approved):
Almost there! 🚀 

I have a few requests regarding documentation and naming, but nothing major. I'll leave it up to you which of those (if any) should be addressed. Approving already!

PR #197, opened 2026-09-16
Feat/point size viewer control

cavenel:

Follows up on the render options added in #191, which exposed `globalPointSizeFactor` only through the project settings dialog.

# Type of change

- [ ] Other

# List of changes made

- Add a `Slider` primitive in `components/ui` on top of `@base-ui/react/slider`, composed like `Switch`: one component that renders track, indicator and one thumb per value.
- Add `PointSizeViewerControl`, a top-right viewer control with a slider and a percent readout for the global point size factor. It is only shown when the project has points.
- The slider position maps to the factor on a log scale from 1% to 500%, with a zero zone at the far left and a snap zone around 100%. The mapping lives in `pointSizeSlider.ts` and is unit tested.
- The factor is rounded to three decimals so the number input in the project settings dialog stays readable.

# Screenshot of the fix

image

# Use of AI

Claude Code was used during the process, and the code was reviewed by a human.

# Testing

- Open a project with points. The slider appears top right; the settings dialog should show the same factor.
- Drag to the far left: points disappear. Drag near the centre: readout snaps to 100%. Far right: 500%.
- Arrow keys on the focused thumb move the value in visible steps.
- `pnpm test` in `apps/tissuumaps` runs the mapping tests.

# Further comments

The settings dialog keeps its number input, so the factor can still be set above 500% by hand. The slider then clamps its thumb to the right end while the readout shows the real value.


apps/tissuumaps/src/components/panels/ViewerPanel/PointSizeViewerControl.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const { globalPointSizeFactor } = glOptions.pointsRenderOptions;
+
+  return (
+    <div
+      className={cn(
+        "m-2 flex items-center gap-2 rounded-xl border border-border bg-background px-3 py-2 shadow-lg",
+        className,
+      )}
+    >
```
jwindhager: I think the div could potentially be replaced by an [InputGroup](https://ui.shadcn.com/docs/components/base/input-group), similar to the seed/shuffle field

jwindhager (changes requested):
Looks (mostly) good to me! I would, however, suggest to inline the functions in pointSizeSlider.ts into the PointSizeViewerControl.tsx, because they are not needed (and shouldn't be re-used) anywhere else. Now that we have MathUtils.remap, I also think we can do without tests for these functions.

apps/tissuumaps/src/components/panels/ViewerPanel/PointSizeViewerControl.tsx
cavenel: For consistency with the drawing div, I would leave it like this.

jwindhager: 👍

cavenel: Actually tested and it doesn't change on the UX... Both are fine with me :) I pushed with the InputGroup.

jwindhager (approved):
Lgtm

PR #199, opened 2026-09-18
fix(render): composite additive images onto an opaque world background owned by the context

jwindhager:

Jira: [TMAP-368 Fix opacity of multi-channel images](https://scilifelab.atlassian.net/browse/TMAP-368)

# Type of change

- [ ] Other

# List of changes made

- **Root cause.** The channels of a multi-channel image are composited additively (`lighter`) onto a black backdrop. `lighter` adds alpha as well as color, and OpenSeadragon clears its canvas to transparent, so an image with opacity below one accumulated alpha from every channel: it darkened towards black and only started fading out once the opacity dropped below `1 / (1 + channels)`. Over an opaque image below it, the compositing was already correct, so the fix is to make the canvas opaque below all content.
- **World background in `OpenSeadragonContext`.** The context now owns an opaque tiled image at world index 0 in a required `backgroundColor`. It spans the content bounds that sources register via `setContentBounds` (each renderer's tiled images, and the points/shapes bounds from the viewer), so it also fits the world, and with it the home view, to all content. Resizes are serialized, as concurrent replacements of the same tiled image would break the index-0 invariant.
- **Renderer anchors.** Anchors now span just the renderer's own tiled images and sit at world index 1 and up; `setExtraBounds` is gone. Newly added tiled images are transformed while hidden, the background is resized, and only then are they shown, so they are never drawn onto a transparent canvas.
- **Viewer wiring.** `Viewer` takes a required `backgroundColor` prop that both constructs the context and styles the viewer element, so canvas and CSS cannot diverge. The app passes black. Changing the color recreates the OpenSeadragon viewer.
- **Pixel URLs.** `OpenSeadragonUtils.createPixelUrl` draws a 1×1 pixel onto a canvas; the transparent and opaque black pixel constants are generated from it. The previous hand-written black pixel PNG had an invalid IDAT checksum that browsers happened to tolerate.
- **Renames.** The context's `updateBounds` is now `updateDummy`, to separate the dummy primitive from the background policy; "extra"/"external" bounds are now "content" bounds throughout.

# Screenshot of the fix

Not relevant.

# Use of AI

Claude Code was used for development.

# Testing

- Open a project with a multi-channel image and lower the image (or layer) opacity: the image fades linearly towards the black background instead of darkening first and fading only at the very end.
- Place a multi-channel image over an RGB image and lower its opacity: it blends towards the image below as before.
- Points or shapes extending beyond the images are still included in the home view, and the navigator still shows the whole world.

# Further comments

The world background assumes the viewer element has the same background color, which `Viewer` now enforces with an inline style. OpenSeadragon's navigator background defaults to black as well.


packages/@tissuumaps-render/src/openseadragon/OpenSeadragonRendererBase.ts
```diff
@@ -892,8 +918,11 @@ export abstract class OpenSeadragonRendererBase<
         } else {
           newRenderedObject.backdrop = backdrop;
           newRenderedObject.tiledImages = tiledImages;
+          // transform first, while still hidden, so that the context's world
+          // background covers the tiled images by the time they are drawn
+          newRenderedObject.pendingBackground = true;
           this._updateRenderedObject(newRenderedObject);
-          await this.updateBounds({ signal });
+          await this._updateBoundsAndReveal([newRenderedObject], { signal });
```
cavenel: With this `await`, `tiledImagesPromise` resolves only after the background is drawn. Two docs still describe the old timing:

- `RenderedObject`: says the promise resolves once the tiled images are in the world.
- `_cleanRenderedObjects`: says waiting for earlier objects "does not delay the additions of this synchronization". It now can, by a frame or two.

The behavior is fine. Please update the two docs.

packages/@tissuumaps-viewer/src/hooks/useOpenSeadragon.ts
```diff
@@ -37,117 +40,124 @@ export function useOpenSeadragon(adapter: ViewerAdapter) {
...
+            await stopOS();
+            return null;
+          })
+          .catch((error) => {
+            console.error("Error stopping OpenSeadragon", error);
+            return null;
+          });
+      };
+    },
+    [backgroundColorHex],
```
cavenel: A change of `backgroundColor` now destroys and recreates the whole viewer. The hex round trip is only there to get a stable dependency. The app passes a constant, so this works today. Option: read the color through a ref, like `osOptionsRef`, or document that the prop is read once.

packages/@tissuumaps-core/src/index.ts
```diff
@@ -32,6 +32,7 @@ export {
   getActiveConfigSource,
   isConstantConfig,
   isFromConfig,
+  findGroupByConfigMap,
```
cavenel: Nit: the export block is alphabetical. This should go before `getActiveConfigSource`.

packages/@tissuumaps-render/src/openseadragon/OpenSeadragonContext.ts
```diff
@@ -64,10 +66,28 @@ export type DataTransfer = {
...
+ * `OpenSeadragonRendererBase.usesAdditiveBlending`). Its color is given on
+ * construction and is expected to match the background behind the viewer,
+ * which shows outside of it. Renderers insert their anchors after it, i.e. at
+ * world index 1 and up. See {@link _resizeBackground} for how it is resized
+ * without ever uncovering the canvas.
  */
 export class OpenSeadragonContext {
   private static readonly _isLittleEndian =
     new Uint8Array(new Uint32Array([1]).buffer)[0] === 1;
+  private static readonly _emptyWorldBounds: Rect = {
```
cavenel: Nit: same unit rectangle as `OpenSeadragonRendererBase._emptyAnchorBounds`. Maybe keep one, e.g. in `OpenSeadragonUtils` next to `hasBounds`.

PR #200, opened 2026-09-19
Feat/tiff labels data provider

cavenel:

Depends on #179 (merged): this branch adds labels on top of the TIFF image data provider.

# Type of change

- [ ] Other

# List of changes made

- Add `TIFFLabelsDataProvider`, `TIFFLabelsData` and `TIFFLabelsDataSource`, so a
  TIFF file can be opened as a label mask. The file, its formats and its pyramid
  are read with the same `openTIFF` as images; the `z` and `t` options pick the
  plane.
- Reject files that cannot hold label IDs: an image drawn in its own colors, more
  than one channel, or samples that are not integers of at most 32 bits. The error
  names what the file actually holds.
- Accept signed as well as unsigned integer masks; `sampleFormatSignedInteger` is
  now named in `TIFFParser`.
- Register the provider in the app.
- Take the label tile data type in `LabelsDataWrapper` from `LabelsData` instead of
  repeating it.
- Document the labels provider on the TIFF supported-data page, with an example
  project of an OME-TIFF plus a CSV-annotated segmentation.

Label IDs are never scanned up front: they are read per tile as the tiles are
drawn, so a large mask opens immediately. TIFF records no label colors, so those
come from the project file, optionally through an annotating `table`.

One commit is marked `feat(storage)!` for widening the tile values to signed
integers, but nothing on `development` depends on that type, so the PR is
additive.

# Screenshot of the fix

# Use of AI

Claude Code was used to transfer OME-Zarr label provider to Tiff. Reviewed by a human.

# Testing

- Unit tests cover the rejection paths of `getLabelLevels`.
- To test in the app: point a project's `labels` at a single-channel integer TIFF,
  as in the example on the TIFF supported-data page.

# Further comments

No


packages/@tissuumaps-storage/src/tiff/TIFFLabelsDataProvider.ts
```diff
@@ -0,0 +1,187 @@
...
+      },
+      t: {
+        type: "integer",
+        minimum: 0,
+      },
+      table: {
+        type: "string",
+      },
+    },
+    required: ["url"], // TODO ... or path
```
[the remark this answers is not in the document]
cavenel: All `path` will be added later, this follows the repo conventions for now.

packages/@tissuumaps-storage/src/tiff/TIFFLabelsDataProvider.ts
```diff
@@ -0,0 +1,187 @@
...
+  const bits = image.getBitsPerSample(0);
+  if (
+    (format !== sampleFormatUnsignedInteger &&
+      format !== sampleFormatSignedInteger) ||
+    bits > 32
+  ) {
+    throw new Error(
+      `The image holds ${bits}-bit samples of TIFF sample format ${format}; label IDs are integers of at most 32 bits.`,
+    );
+  }
```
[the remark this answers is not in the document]
cavenel: Correct that TIFF allows BitsPerSample and SampleFormat to differ between IFDs.

But the whole TIFF module checks level 0 only. `TIFFUtils.validatePyramids` reads
samples per pixel and photometric interpretation from `pyramid[0]`. The
full-range-8-bit check in `TIFFImageDataProvider` reads level 0. This function
does the same.

No writer produces such a file, since pyramid levels downsample the same
samples. Checking every level here would make labels stricter than images. If we
close the hole, we close it in `validatePyramids`, so images and labels both get
it. Leaving it as is.

apps/docs/docs/supported-data/tiff.md
```diff
@@ -47,9 +48,15 @@ Settings without a fallback are reported as unset, leaving the renderer's defaul
 
 All of these can be overridden per channel in the project file through the image's `channels` array (see the [example](#example) below).
 
+## Labels
+
+TIFF files are also opened as **labels**, where every pixel value is a label (segment) ID and `0` is background. The formats and the pyramid are read as for images, with two restrictions: the file has to hold a **single channel** (an RGB file, or a multi-channel file, is rejected), and its samples have to be **integers of at most 32 bits** (signed or unsigned).
```
jwindhager: "also"? Perhaps rephrase to "can also be"

jwindhager: It would be good to state somewhere that one has to be careful with downsampling when generating pyramidal labels.

apps/tissuumaps/src/data/cache/wrappers/LabelsDataWrapper.ts
```diff
@@ -20,11 +18,9 @@ export class LabelsDataWrapper
...
   }
 
-  getTileData(event: Parameters<LabelsData["getTileData"]>[0]): Promise<{
-    values: number[] | IntArray | UintArray;
-    width: number;
-    height: number;
-  }> {
+  getTileData(
+    event: Parameters<LabelsData["getTileData"]>[0],
+  ): ReturnType<LabelsData["getTileData"]> {
```
jwindhager, line 23: Why update the return type? Using Parameters<> for the event type is necessary to avoid having to import OpenSeadragon here, but I think we should prefer explicit types for catching errors/inconsistencies in case we ever update the base type.

jwindhager (approved):
Thanks, awesome!

Only minor comments, resolve at your discretion

apps/tissuumaps/src/data/cache/wrappers/LabelsDataWrapper.ts
cavenel: Yeah I actually removed it on purpose! As the wrapper only forwards the call:
```
getTileData(event) {
  return this.data.getTileData(event);
}
```
If `LabelsData.getTileData` changes its return type, that line would break. Plus writing the shape out a second time adds no check: TypeScript already compares the two at the delegate call.

The sibling wrapper already does it this way. `ImageDataWrapper.ts:20-26` types every member as `ImageData["getTileData"]` etc. Deriving from the interface is the existing convention. The event parameter here already used `Parameters<>;` this just makes the return match.

What do you think? Happy to revert to the OMEZarr way, or to fix OMEZarr in the same way, up to you!

apps/docs/docs/supported-data/tiff.md
cavenel: Added!

cavenel: Fixed!

apps/tissuumaps/src/data/cache/wrappers/LabelsDataWrapper.ts
jwindhager: Ah, sorry, completely missed that this was just the wrapper. Good change if course, sorry again!

PR #201, opened 2026-09-19
feat: single `source` field for data sources

jwindhager:

Touches upon several issues, but none directly & comprehensively describe this refactoring

# Type of change

- [ ] Other

# List of changes made

**Breaking: data sources reference their data through a single `source` string.** The `url` and `path` fields are removed without migration; existing project files have to be updated.

- `source` is one of: a URL (`https://…`, `blob:…`, `data:…`), an app-relative path (`//data/x.csv`), a workspace-relative path (`/data/x.csv`), or a project-relative path (`data/x.csv`, `./data/x.csv`, `../shared/x.csv`). Project-relative paths resolve against where the project was loaded from; without a project source they fall back to workspace-relative, and without an open workspace to app-relative.
- Two former URL forms change meaning: an origin-absolute `/data/x.csv` is now a workspace path, and a protocol-relative `//host/x.csv` is now an app-relative path. A file name whose first segment contains a colon has to be written as `./name:x.tif`.
- New `SourceUtils` in `@tissuumaps/core`: `normalizeSource()` classifies a source, applies the fallbacks and normalizes it synchronously to an absolute URL or a workspace-relative path; `resolveSource()` opens a normalized workspace path as a `FileSystemFileHandle`. Unit-tested, including with a fake directory-handle tree.
- `DataProvider.normalize()` now receives the workspace handle and the project source instead of the project URL. All built-in providers (CSV, Parquet, GeoJSON, OpenSeadragon, OME-Zarr images and labels, TIFF) expose `source` in their schemas, normalize it with `SourceUtils.normalizeSource()` and resolve it in `load()` with `SourceUtils.resolveSource()`. `source` is required on their data source interfaces; it stays optional on the base `RawDataSource` for table-backed points and for OpenSeadragon's `tileSourceConfig` alternative.
- The project store's `url` is renamed to `source` and can hold a workspace-relative project path. `DataCache` normalizes with the workspace and project source, memoizes on both, and treats an entry as workspace-dependent when its normalized source is a workspace path.
- `loadProjectFromFile()` takes a `File` and loads without a source; new `loadProjectFromWorkspace()` takes a `FileSystemFileHandle` and the workspace handle and loads with the file's workspace-relative path (not wired to the UI yet).
- Docs: the "Referencing data" section describes the new syntax and fallbacks; the OME-Zarr and TIFF pages and the supported-data table use `source`.

# Screenshot of the fix

Not applicable.

# Use of AI

Claude Code was used during development for reviewing, implementing and refactoring the changes, writing tests and docs, and drafting this description; all changes were reviewed and directed by the author.

# Testing

- Type-check with `pnpm build`, lint with `pnpm lint`, and run `pnpm test`: the `SourceUtils` and `DataCache` tests cover normalization, fallbacks, idempotence, workspace resolution and cache re-normalization.
- Manual: open a project with `source` fields from a URL and verify project-relative sources resolve next to the project file. Workspace resolution has no UI entry point yet and is covered by unit tests only.

# Further comments

- The prefixes were chosen deliberately: bare paths keep their URL-like project-relative meaning, `/` marks the workspace, `//` the app.
- `SourceUtils` assumes well-formed project sources and normalized sources; validation happens in `normalizeSource()`.


jwindhager:
@HenrikeW I realized the url/path split was very confusing to both users and developers; this PR should simplify things

packages/@tissuumaps-core/src/utils/SourceUtils.ts
```diff
@@ -0,0 +1,367 @@
...
+ * segments and fails if the path leaves the workspace.
+ *
+ * A relative path whose first segment contains a colon (e.g.
+ * `sample1:ch2.tif`) is taken for a URL; write it as `./sample1:ch2.tif`
+ * instead.
+ */
+export class SourceUtils {
+  private static readonly _pathSep = "/";
+  private static readonly _appPathPrefix = "//";
+  private static readonly _workspacePathPrefix = SourceUtils._pathSep;
```
jwindhager: Should not use _pathSep here.

packages/@tissuumaps-core/src/utils/SourceUtils.ts
```diff
@@ -0,0 +1,367 @@
...
+ * against a URL follows URL semantics: `..` segments beyond the root of the
+ * URL's path are dropped silently. Resolving within the workspace collapses the
+ * segments and fails if the path leaves the workspace.
+ *
+ * A relative path whose first segment contains a colon (e.g.
+ * `sample1:ch2.tif`) is taken for a URL; write it as `./sample1:ch2.tif`
+ * instead.
+ */
+export class SourceUtils {
+  private static readonly _pathSep = "/";
```
jwindhager: No need to make this a configurable constant

packages/@tissuumaps-core/src/utils/SourceUtils.ts
```diff
@@ -0,0 +1,367 @@
...
+    if (projectSource !== null) {
+      if (SourceUtils._urlSchemePattern.test(projectSource)) {
+        if (!URL.canParse(projectPath, projectSource)) {
+          throw new Error(`Invalid project-relative path: ${projectPath}`);
+        }
+        return new URL(projectPath, projectSource).toString();
+      }
+      if (workspace === null) {
+        throw new Error(
+          `Cannot normalize project-relative path without workspace: ${projectPath}`,
```
cavenel: `normalize()` runs synchronously in `retainOnly()` inside the store subscribers, with no try/catch. Anything `normalizeSource` throws escapes `setState`, leaving the store updated but the caches half-reconciled.

This error is user-triggered: close the workspace while a project loaded from it is open, and every project-relative source ends up here. The new `""` and `..`-escape throws reach the same place from a malformed project file, where a bad `path` used to fail per-object at load time.

Either fall back like the other missing-context cases (workspace-relative, hence app-relative), or catch per data source in `_resolveDataSource` and let the entry's load operation fail instead.

apps/tissuumaps/src/data/cache/DataCache.ts
```diff
@@ -490,18 +500,24 @@ export class DataCache<
     if (
       cached !== undefined &&
       cached.dataProvider === dataProvider &&
-      cached.projectUrl === context.projectUrl
+      cached.workspace === context.workspace &&
```
cavenel: The doc block above `_resolveDataSource` is now stale: it still says the memo is recomputed "whenever the data provider registered for the data source's type or the project URL changes - both of which normalization depends on". It is three inputs now: the data provider, the workspace and the project source. (`code-architecture.md` was updated for exactly this.)

packages/@tissuumaps-core/src/utils/SourceUtils.ts
```diff
@@ -0,0 +1,367 @@
...
+   * @param options - Optional base URL that the path is resolved against when
+   *   falling back to being app-relative (default `document.baseURI`)
+   * @returns The absolute URL, or the normalized workspace-relative path
+   * @throws Error if the path does not form a valid URL with the project URL,
+   *   if the project source is workspace-relative but no workspace is open, if
+   *   the path leaves the workspace or does not name a file, or if it cannot be
+   *   normalized as a workspace-relative path (see
+   *   {@link SourceUtils.normalizeWorkspacePath})
+   */
+  static normalizeProjectPath(
```
cavenel: `normalizeProjectPath`, `normalizeWorkspacePath` and `normalizeAppPath` have no caller outside this class - everything goes through `normalizeSource`, `resolveSource`, `isWorkspacePath` and `makeWorkspacePath`. `JSONUtils` keeps its helpers `private static _…`; doing the same here halves the exported surface and the `@throws` chains that cross-reference each other.

The two "throws if the prefix is missing" tests go with them - those checks are unreachable through `normalizeSource`, which only dispatches once the prefix is there.

packages/@tissuumaps-core/src/utils/SourceUtils.ts
```diff
@@ -0,0 +1,367 @@
...
+   *   {@link SourceUtils.normalizeProjectPath}), a workspace-relative path
+   *   cannot be normalized (see {@link SourceUtils.normalizeWorkspacePath}),
+   *   or an app-relative path cannot be normalized (see
+   *   {@link SourceUtils.normalizeAppPath})
+   */
+  static normalizeSource(
+    source: string,
+    workspace: FileSystemDirectoryHandle | null,
+    projectSource: string | null,
+    options?: { baseUrl?: string },
```
cavenel: `options.baseUrl` is test-only - no production call site passes it, yet it is threaded through four signatures and documented in each. The suite already stubs `document.baseURI` with `vi.stubGlobal` in "defaults to the document base URL"; doing that everywhere lets the option go.

packages/@tissuumaps-core/src/utils/SourceUtils.ts
cavenel: `normalize()` runs synchronously in `retainOnly()`, inside the store subscribers, with no try/catch. So anything `normalizeSource` throws escapes `setState`, and the store ends up updated while the caches are only half reconciled.

A user can trigger this one: close the workspace while a project loaded from it is open, and every project-relative source lands here. The new throws for `""` and for a `..` that escapes reach the same place from a malformed project file, where a bad `path` used to fail per object at load time.

You could fall back like the other missing-context cases, or catch per data source in `_resolveDataSource` and let the entry's load operation fail instead.

apps/tissuumaps/src/data/cache/DataCache.ts
cavenel, line 524: The doc block above `_resolveDataSource` is outdated now. It still names the data provider and the project URL as the two inputs of normalization. There are three now: the data provider, the workspace and the project source. (`code-architecture.md` was updated for exactly this)

packages/@tissuumaps-core/src/utils/SourceUtils.ts
cavenel: `normalizeProjectPath`, `normalizeWorkspacePath` and `normalizeAppPath` have no caller outside this class. Everything goes through `normalizeSource`, `resolveSource`, `isWorkspacePath` and `makeWorkspacePath`. `JSONUtils` keeps its helpers `private static _…`, and the same here would halve the exported surface and the `@throws` chains that cross reference each other.

The two "throws if the prefix is missing" tests go with them. Those checks are unreachable through `normalizeSource`, which only dispatches once the prefix is there.

packages/@tissuumaps-core/src/utils/SourceUtils.ts
cavenel, line 105: Only the tests pass `baseUrl`. The app never does. Will we need it in the future?

cavenel (commented):
Super nice!!
Very minor comments, but then can be merged!

packages/@tissuumaps-core/src/utils/SourceUtils.ts
jwindhager: This is test-only API surface - without it, testing it would require stubbing a DOM global (tests in @tissuumaps/core currently run in node, not jsdom). I don't think its harmful/wrong to keep; will add a comment about it to the TSDoc

packages/@tissuumaps-core/src/utils/SourceUtils.ts
jwindhager: This actually were two independent bugs: (1) the fallback didn't work properly, and (2) normalization errors weren't caught. Both are solved now.

apps/tissuumaps/src/data/cache/DataCache.ts
jwindhager: Done

packages/@tissuumaps-core/src/utils/SourceUtils.ts
jwindhager: Private now; tests gone.

jwindhager:
Merging. Would you mind updating the example data, @cavenel?

